PR #16, opened 2015-12-25
Added fingerprint plugin to the list.

smstuebe:


smstuebe:
I'll create a new one for the changed name space

PR #46, opened 2016-07-04
Estimote SDK

Redth:
Moving bindings to Android/iOS Estimote SDK's into public repo.  This also contains updated bindings to the latest versions of each platforms' SDK.


afgallo:
@Redth thank God I saw this ! I've been planning on using Xamarin to write an app which integrates with estimote for a couple of months now but the bindings were way too old!
Is there any plan to publish this into nuget any time soon?

Thanks so much for this!

PR #56, opened 2016-08-07
added bluetooth LE plugin

smstuebe:


jamesmontemagno:
I have added In alan's BLE one for now.

PR #57, opened 2016-08-14
Google Project Tango

mattleibow:
I have done the binding for https://developers.google.com/tango/

It is almost there, but could use a little more love.
# Notes

Our binding tools cannot fully parse the SDK jar due to the fact that it is referencing some private type from another jar: `ITangoCloudService` and `ITangoCloudService.Stub`. This results in major parts of the binding being skipped. To work around this, I created a dummy jar which has this interface. It is set as a pure `ReferenceJar` and is not included at all. This should not be any problem for consumers since both this type is private and a native Java application doesn't have any issues with the "missing" type.

There is an issue with downloading the jar and aar files from Google as the user needs to be logged in and accept some T&Cs. We need to fins a way to work around this.
# Done & To Do
- The binding is done, but could use love and .NET-ifying
  - changing consts into enums
  - const names need cleaning
  - eg: `KeyStringAreadescription` should be `KeyStringAreaDescription`
  - eg: `TangoConfig.ConfigTypeAreaDescription` could be `TangoConfigTypes.AreaDescription`
- The samples need to be created
  - right now there is just the one: Basic Examples - HelloAreaDescription
  - more Java: https://github.com/googlesamples/tango-examples-java
- The component needs to be doc-ified
  - check the licensing
  - create the details and getting started guides
- The NuGet is done
- The icons are created


newky2k:
Project tango is now abandoned

newky2k:
Project tango is now abandoned

PR #72, opened 2016-09-30
[card.io] updated to latest versions of native SDK's for iOS and Android

Redth:


bholmes:
@Redth  component.yaml for iOS need a version update.

PR #83, opened 2016-10-28
Estimote iOS and Android versions updated

adlair:


bholmes:
Please resubmit without Dropbox files being changed

PR #86, opened 2016-11-14
Adding Jackson Core for Android

bholmes:


Redth (commented):
Should we come up with some sort of README for this?

bholmes:
@Redth added the readme.  really simple.  suggestions welcome

Redth:
I don't think it needs to be more elaborate than that, just nice to have _something_ to land on.

Android/Jackson.Core/nuget/Jackson.Core.nuspec
```diff
@@ -0,0 +1,19 @@
...
+    <authors>Xamarin Inc.</authors>
+    <owners>Xamarin Inc.</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>This project contains core low-level incremental ("streaming") parser and generator abstractions used by Jackson Data Processor.</description>
+    <copyright>Copyright 2016 Xamarin Inc</copyright>
+    <projectUrl>https://github.com/xamarin/XamarinComponents/tree/master/Android/Jackson.Core</projectUrl>
+    <licenseUrl>https://github.com/xamarin/XamarinComponents/blob/master/Android/Jackson.Core/License.md</licenseUrl>
+  </metadata>
+  <files>  
+    <file src="output/Jackson.Core.dll" target="lib/MonoAndroid23" />
```
Redth, line 16: I couldn't find any mention of lowest supported android version on the source repo, however I'm guessing since it's really not android specific code, this should be fine.  Has it been tested on 2.3 at all?

bholmes: TBH I am guessing.  This is not an Android library from what [I can tell.](https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-core/2.8.5/jackson-core-2.8.5.pom)  Also I do not get any more (or less) warnings from classparse when the api level is changed.

PR #111, opened 2017-01-18
[AdvancedColorPicker] Added the component

mattleibow:


iOS/AdvancedColorPicker/source/AdvancedColorPicker/ColorPickerView.cs
```diff
@@ -0,0 +1,218 @@
...
+		public UIColor SelectedColor
+		{
+			get
+			{
+				return UIColor.FromHSB(satbrightPicker.Hue, satbrightPicker.Saturation, satbrightPicker.Brightness);
+			}
+			set
+			{
+				nfloat hue = 0, brightness = 0, saturation = 0, alpha = 0;
+				value.GetHSBA(out hue, out saturation, out brightness, out alpha);
```
rmarinho: Yo need to check for null here..

rmarinho (changes requested):

mattleibow:
@rmarinho I updated it. If the property is set to null, the the color values will all be 0 => black

PR #128, opened 2017-03-13
Add option to overwrite the source assembly in XBD

bholmes:
With Android builds we are seeing a number of issues

 - Warnings that assemblies referenced from the NuGet can not be found
   (Not errors from what I can tell.)
 - Build time errors on first build because resources can not be found
   (Clean builds appear to fix this too, but yuck)
 - Slow VS experience when IDE is at idle.  We believe this is because of
   design time builds that are running.  Changing the reference list after
   references are resolved causes more processes to spawn.

 This patch avoids changing the reference list after the resolve assembly step runs.
 Instead the source assembly is updated and written back to the same location and not
 the intermediate directory.  It appears to work in the cases I tested but we need to
 run a full set and have the community tell us how it is working.

 There is now a property on all of the resource restore tasks that allow the behavior to be configured.
 On Android we default to overwrite and on iOS we continue to use the intermediate folder by default.

 We need to add an optimization to know if we need to open the assembly in cecil.  Right now we always
 open the file.  It is taking a second in my tests.  We could write out and check for a 'marker' file.

 This will buy us some more time as a temporary workaround.  We are working on changes
 needed in Xamarin Android so that we can add aar files to the project build list and not
 require that we re-embed the aar or reference files.

### DIFF
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/BaseXamarinBuildResourceRestore.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/BaseXamarinBuildResourceRestore.cs
@@ -25,6 +25,8 @@ namespace Xamarin.Build.Download
 
 		public bool ThrowOnMissingAssembly { get; set; } = true;
 
+		public virtual bool  OverwiteSourceAssembly { get; set; } = false;
+
 		[Output]
 		public ITaskItem [] ChangedReferencePaths { get; set; }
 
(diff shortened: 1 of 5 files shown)


bholmes:
Keep an eye on the following when reviewing this PR.

[bug 52967](https://bugzilla.xamarin.com/show_bug.cgi?id=52967)
[Forums Post](https://forums.xamarin.com/discussion/comment/259527#Comment_259527)

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/BaseXamarinBuildResourceRestore.cs
```diff
@@ -104,9 +104,11 @@ public override bool Execute ()
 		bool MergeResources (IAssemblyResolver resolver, string originalAsmPath, string mergedAsmPath, string assemblyName, List<ITaskItem> resourceItems)
 		{
 			var disposeList = new List<IDisposable> ();
```
kentcb, line 124: Crying out for some `CompositeDisposable` goodness :)

bholmes: Nice suggestion!  I would like to use that however I just found that it requires that I ship additional dlls in the NuGet just for a simple concept.

PR #144, opened 2017-04-21
[Mono.Options] Added the component to the public repo and updated

mattleibow:
Now using the latest from mono: v5.3.0

I also added .NET Standard 1.3 support, which provides better features for all platforms. Originally, there was `net4` and `portable-`. The PCL build had a few features removed, but not too much. The .NET Standard build has a few of them back, so on iOS/Android the .NET Standard assembly is used instead of the PCL assembly. The PCL is still used from PCL projects or Win8/WP8.

For best features, a .NET Standard project should be used. As soon as .NET Standard 2.0 becomes a stable thing, I will add that and we can get the "full" feature set (serialization/type converters) to all apps.

This uses the new bits from this commit: https://github.com/mono/mono/commit/c1b9fc566a365cf6923218eae9098d0fd0e632ae
With changes in this PR: https://github.com/mono/mono/pull/4719
Merged here: https://github.com/mono/mono/commit/7e2571ed334e9cee3f0d3bafeef02852310f4d3b


mattleibow:
@bholmes I have also changed the licensing from the private repo. 
 - License.md: this is the typical, short .NET Foundation license
 - External-Dependency-Info.txt: this includes the "components header" with the full mono license

The mono license has a few licenses :) And, it mentions patents. I am not sure exactly how to handle that - or if I can just let it be.

bholmes (approved):

PR #158, opened 2017-05-23
Update the READMEs for the Firebase Jobdispatcher binding

topgenorth:
Forgot to update the README for this, which provides a link to docs on Xamarin developer portal.

PR #160, opened 2017-05-24
The NuGet was missing the reference to the Support v4 Nuget.

topgenorth:
This replaces PR #158 - Updated the README's for the binding.


bholmes (approved):
LGTM

Android/FirebaseJobDispatcher/nuget/Xamarin.Firebase.JobDispatcher.nuspec
```diff
@@ -11,6 +11,9 @@
     <copyright>Copyright (c) 2017 Xamarin Inc</copyright>
     <projectUrl>https://github.com/xamarin/XamarinComponents/tree/master/Android/FirebaseJobDispatcher</projectUrl>
     <licenseUrl>https://github.com/xamarin/XamarinComponents/blob/master/Android/FirebaseJobDispatcher/License.md</licenseUrl>
+    <dependencies>
+        <dependency id="Xamarin.Android.Support.v4" version="25.3.1" />
```
Redth, line 15: Shouldn't `Xamarin.GooglePlayServices.Gcm` be listed as a dependency too? or is it not strictly _required_

topgenorth:
So AFAIK, Google Play Services has to be installed on the device, but GCM itself is not a dependency of the library (the Firebase Job Dispatcher docs are kind of hazy on this).  It's kind of weird: You can install an app that uses the FJD on a device without Google Play Services and it will run - however scheduling jobs silently fails (or that is what happens in my tests).

Redth:
So, curious why you have them as a reference on the project?

topgenorth:
So, it seems that the binding will work without having the GCM reference (at least it did on my device). The existing GCM references were a left-over from when I first started on the binding and GCM was necessary. 

I've updated the binding project by removing GCM, and consolidated to a single solution (instead of one for the sample and one for the binding). I'll open another PR for those changes, I believe we can close this issue off.

PR #165, opened 2017-06-07
Remove GCM dependency from Xamarin.Firebase.JobDispatcher

topgenorth:
After a bit of experimenting, it seems that the GCM dependency isn't necessary in the binding project. Removed the dependency. I also consolidated the binding & sample projects into one `SLN` file (before we had one `SLN` for the binding and another for the sample).


Android/FirebaseJobDispatcher/source/Firebase.JobDispatcher/packages.config
```diff
@@ -8,9 +8,6 @@
   <package id="Xamarin.Android.Support.Media.Compat" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Android.Support.v4" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Build.Download" version="0.4.5" targetFramework="monoandroid71" />
-  <package id="Xamarin.GooglePlayServices.Base" version="42.1021.1" targetFramework="monoandroid71" />
   <package id="Xamarin.GooglePlayServices.Basement" version="42.1021.1" targetFramework="monoandroid71" />
-  <package id="Xamarin.GooglePlayServices.Gcm" version="42.1021.1" targetFramework="monoandroid71" />
-  <package id="Xamarin.GooglePlayServices.Iid" version="42.1021.1" targetFramework="monoandroid71" />
   <package id="Xamarin.GooglePlayServices.Tasks" version="42.1021.1" targetFramework="monoandroid71" />
```
bholmes: I think we may have covered this before.  Apologies if that is the case.  

Do we need Xamarin.GooglePlayServices.Tasks and Xamarin.GooglePlayServices.Basement here?  If so then they should be added to the nuspec as a dependency and the sample nuget list to be complete??

bholmes (commented):
I was able to remove the Xamarin.Android.Support.v7.CardView package from the sample and the project still built.  Not sure if it runs.  You may want to double check that one.

Android/FirebaseJobDispatcher/source/Firebase.JobDispatcher/packages.config
topgenorth: CardView is required - the sample app won't run without it. I did manage to remove `Xamarin.GooglePlayServices.Tasks` and `Xamarin.GooglePlayServices.Basement` without any ill effect observed in the sample app. Not to sure what dragged those dependencies in.

My lastest commit reflects these changes.

bholmes (approved):

Android/FirebaseJobDispatcher/source/Firebase.JobDispatcher/packages.config
Redth: Gcm/Iid would have brought in Tasks and Basement so since you removed those, the others are safe to go.

Redth (approved):

PR #166, opened 2017-06-08
[XBD] Manually Extract proguard configs from .aar files

Redth:
Currently `proguard.txt` files were _not_ being picked up from inside of .aar files (really any `LibraryProjectZip` item) by the proguard build task.

In this patch we are going to manually spot them and extract them to the same directory as we output the assembly with the merged resources to so they can be added to `@(ProguardConfiguration)` which the proguard build task will then see.

### DIFF
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
@@ -43,8 +43,6 @@ namespace Xamarin.Build.Download
 			if (restoreMap == null)
 				return false;
 
-			IAssemblyResolver resolver = CreateAssemblyResolver ();
-
 			// Look through all the assemblies we would restore for
 			foreach (var asm in restoreMap) {
 
@@ -94,6 +92,9 @@ namespace Xamarin.Build.Download
 							// Figure out our destination filename
 							var proguardSaveFilename = Path.Combine (proguardIntermediateOutputPath, saveNameHash + entryCount + ".txt");
 
+							// Add this to our file writes
+							additionalFileWrites.Add (new TaskItem (proguardSaveFilename));
+
 							// Save out the proguard file
 							using (var entryStream = entry.Open ())
 							using (var fs = File.Create (proguardSaveFilename)) {
@@ -105,8 +106,6 @@ namespace Xamarin.Build.Download
 							entryCount++;
 						}
 					}
-
-					entryCount++;
 				}
 
 				// *.proguard.stamp files are additional file writes
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
@@ -20,6 +20,11 @@ namespace Xamarin.Build.Download
 
 		string proguardIntermediateOutputPath = null;
 
+		// We intentionally won't call the base implementation in this override
+		// since other tasks should handle the restores
+		// This task is just responsible for extracting proguard config files
+		// from the .aar input files so we are reusing the base task to help
+		// track down the .aar files themselves.
 		public override bool Execute ()
 		{
 			// Get the dir to store proguard config files in
@@ -27,20 +32,24 @@ namespace Xamarin.Build.Download
 			if (!Directory.Exists (proguardIntermediateOutputPath))
 				Directory.CreateDirectory (proguardIntermediateOutputPath);
 
-
 			var additionalFileWrites = new List<ITaskItem> ();
 
+			// Make sure our XbdMerge directory exists
 			var outputDir = MergeOutputDir;
 			Directory.CreateDirectory (outputDir);
 
-			var restoreMap = base.BuildRestoreMap (RestoreAssemblyResources);
-			if (restoreMap == null) {
+			// Get our assembly restore map
+			var restoreMap = BuildRestoreMap (RestoreAssemblyResources);
+			if (restoreMap == null)
 				return false;
-			}
 
 			IAssemblyResolver resolver = CreateAssemblyResolver ();
 
+			// Look through all the assemblies we would restore for
 			foreach (var asm in restoreMap) {
+
+				// We only want to find proguard files in .aar files referenced
+				// for assemblies we actually have referenced and have them mapped to
 				var asmName = new AssemblyName (asm.Key);
 				ITaskItem item = FindMatchingAssembly (InputReferencePaths, asmName);
 				if (item == null) {
@@ -50,19 +59,22 @@ namespace Xamarin.Build.Download
 						continue;
 				}
 
-				var intermediateAsmPath = Path.Combine (outputDir, asmName.Name + ".dll");
-
+				// Use a hash for the assembly name to keep paths shorter
 				var saveNameHash = DownloadUtils.HashMd5 (asmName.Name)?.Substring (0, 8);
 
-				// Keep a stamp file around to avoid reprocessing
+				// We keep a stamp file around to avoid reprocessing, so skip if it exists
 				var stampPath = Path.Combine (outputDir, saveNameHash + ".proguard.stamp");
 				if (File.Exists (stampPath))
 					continue;
 
+				// Get all the mapped .aar files
 				var resourceItems = asm.Value;
 
+				// We want to increment on the hash name in case there are multiple .aar files and/or proguard config
+				// files for a given assembly, so we use them all and not overwrite the same name
 				var entryCount = 0;
 
+				// In theory we could have multiple .aar files? Probably never happen...
 				foreach (var resourceItem in resourceItems) {
 					
 					// Full path to .aar file
@@ -97,6 +109,7 @@ namespace Xamarin.Build.Download
 					entryCount++;
 				}
 
+				// *.proguard.stamp files are additional file writes
 				File.WriteAllText (stampPath, string.Empty);
 				additionalFileWrites.Add (new TaskItem (stampPath));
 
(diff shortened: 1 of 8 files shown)


bholmes (commented):
I am concerned writing back to the merged asm dir.  IIRC this is the packages/lib/target folder.  I am wondering what will happen if and when the NuGet client and MSBuild tasks stop adding NuGet dlls as references to the csproj and search the directory for content.  Can having an unknown file in this folder cause bugs for future build engine updates?  I may just be overthinking it.  Otherwise I feel the PR does what it intends to do and LGTM.

Redth:
I wonder if @mrward would have any comment on this...

mrward:
Not sure. Where are these new files being created? Into the packages directory itself? So in a PackageReference world these new files would be generated into the ~/.nuget/packages/ directory for the corresponding NuGet package?

Redth:
They would end up going into the local sln dir's `./packages/Xamarin.GooglePlayServices.Basement/lib/MonoAndroid70/` for example, and the file would be `proguard.txt`

We have a bit of history around the process here, but essentially we modify the .dll files in that same path (in place) to work around thrashing of References in the .csproj causing issues (like extra designer build invocations).  This part seems to work well.

We're just talking in this case about adding a .txt to that libs folder.

mrward:
In a PackageReference world there is no $(SolutionDir)/packages so these files I guess would be extracted to ~/.nuget/packages/Xamarin.GooglePlayServices.Basement/lib/MonoAndroid70 and would be available to all projects that use that NuGet package. If these are just .txt and .cfg files then this should be fine.

I would be tempted to just create an Android project, and add a PackageReference, and try a restore and a build to see what happens. Just to test things.

```
<ItemGroup>
   <PackageReference Include="Xamarin.GooglePlayServices.Basement" Version="1.2.3.4" />
</ItemGroup>
```

Then from the command line try `msbuild /t:restore YourSoluton.sln` and an `msbuild /t:Build YourSolution.sln` and see if that works. The restore will generate a few files in the obj/ directory. The project.assets.json has the assemblies that will be referenced, which I guess will not be a problem here.

Looking at the changes I cannot see any problems. Just adding a ProguardConfiguration task item for the extracted file looks fine to me. I guess adding your own proguard config file to customize things still works here.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarRestore.cs
```diff
@@ -4,19 +4,53 @@
...
+			if (!Directory.Exists (proguardIntermediateOutputPath))
+				Directory.CreateDirectory (proguardIntermediateOutputPath);
+			
+			var res = base.Execute ();
+
+			Log.LogMessage ("Looking for proguard configuration files in: {0}", proguardIntermediateOutputPath);
+
+			// Find any proguard config files we need to emit that were extracted
+			AarProguardConfiguration = Directory.GetFiles (proguardIntermediateOutputPath)
+			                                    ?.Select (pf => new TaskItem(pf))?.ToArray ()
```
bholmes: Add Where for extension

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarRestore.cs
```diff
@@ -4,19 +4,53 @@
...
 	{
 		public bool FixAndroidManifests { get; set; } = true;
 
-		protected override Stream LoadResource (string resourceFullPath, string assemblyName)
+		[Output]
+		public ITaskItem [] AarProguardConfiguration { get; set; }
+
+		string proguardIntermediateOutputPath = null;
+
+		public override bool Execute ()
```
bholmes: Check switching from Debug to Release.  Wondering if the LoadResource call will be made as the modified dll is already written.  The base may skip in this case.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarRestore.cs
```diff
@@ -4,19 +4,53 @@
...
+			                                    ?.Select (pf => new TaskItem(pf))?.ToArray ()
+			                                        ?? new ITaskItem[] {};
+
+			foreach (var apc in AarProguardConfiguration)
+				Log.LogMessage ("ProguardConfiguration: {0}", apc);
+			
+			return res;
+		}
+
+		protected override Stream LoadResource (string resourceFullPath, string assemblyName, string assemblyOutputPath)
```
bholmes: assemblyOutputPath is no longer being used.  consider reverting it.

bholmes (dismissed):

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarRestore.cs
Redth: So I think this is actually the case, the base will skip this. I think I'm going to need to move this bit into its own task that is always run.

Redth:
Reworked this entire PR and pulled out the proguard config extraction into its own task.  I was running into a few cases where the task would be skipped too early to extract the proguard configs in all the correct scenarios when trying to do this as part of the existing aar restore task.

The big one was we checked for the assembly resource for the library project zip and if it was already embedded, we'd skip the rest of the task, so the config files wouldn't get extracted again if the user ran a clean, but after the initial .aar restore had happened and directly overwrote the packages/*.dll files.

I think this also makes the code much easier to remove once this is fixed upstream in Xamarin.Android's core tasks.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
```diff
@@ -0,0 +1,123 @@
...
+
+					using (var fileStream = File.OpenRead (resourceFullPath))
+					using (var zipArchive = new ZipArchive (fileStream, ZipArchiveMode.Read)) {
+
+						// Look for proguard config files in the archive
+						foreach (var entry in zipArchive.Entries) {
+							
+							// Skip entries which are not proguard configs
+							if (!entry.Name.Equals ("proguard.txt", StringComparison.OrdinalIgnoreCase)
+								&& !entry.Name.Equals ("proguard.cfg", StringComparison.OrdinalIgnoreCase))
```
mattleibow, line 89: Just a question/note: are these the only names that are used by Google? Is this name arbitrary, or does the .aar have a specific convention and thus we can assume this?

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
```diff
@@ -0,0 +1,123 @@
...
+								continue;
+
+							// Figure out our destination filename
+							var proguardSaveFilename = Path.Combine (proguardIntermediateOutputPath, saveNameHash + entryCount + ".txt");
```
mattleibow, line 93: Should we be adding this path to the AdditionalFileWrites?

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
```diff
@@ -0,0 +1,123 @@
...
+			// Make sure our XbdMerge directory exists
+			var outputDir = MergeOutputDir;
+			Directory.CreateDirectory (outputDir);
+
+			// Get our assembly restore map
+			var restoreMap = BuildRestoreMap (RestoreAssemblyResources);
+			if (restoreMap == null)
+				return false;
+
+			IAssemblyResolver resolver = CreateAssemblyResolver ();
```
mattleibow: Are we using this anywhere? This looks like it may be hitting the file system?

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
```diff
@@ -0,0 +1,123 @@
...
+								entryStream.CopyTo (fs);
+								fs.Flush ();
+								fs.Close ();
+							}
+
+							entryCount++;
+						}
+					}
+
+					entryCount++;
```
mattleibow: I may be nitpicking, but de we need to increment here? we are basically going through a collection of collections, and if we are incrementing each time in the inner loop, this technically will skip a number. I am just thinking that if we don't have to increment, then don't - we may be debugging this later and we may think that something is wrong when our files go 1, 2, 4, 5, 7, 8 (or something like that).

mattleibow (commented):
@Redth this looks good. I have mostly nitpicking questions, except for maybe the AdditionalFileWrites one.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
Redth: 👍 forgot to remove.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
Redth: Yeah so from what i've seen proguard.txt is the convention.  proguard.cfg is something we use so i've thrown that in there on the assumption it may occur in .aar's as well, but i sort of doubt it.  In any case i don't think there's harm in adding proguard.cfg too since the chances of some other random file having that name with some other meaning are next to none.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
Redth: Yes.  It does get cleaned out regardless with the custom clean task, but it would be good practice to add these, will change.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildAndroidAarProguardConfigs.cs
Redth: Yep, you're right.  I was guarding incorrectly against multiple .aar files but hadn't quite fully contemplated the logic :)

Redth (commented):
Fixed proposed changes.

PR #167, opened 2017-06-12
Added Google.Android.Vending

mattleibow:
These are bindings for the Google Play LVL and APK Expansions:

 - https://github.com/google/play-licensing
 - https://github.com/google/play-apk-expansion

This is a new binding library to replace my original C# port: https://github.com/mattleibow/Android.Play.ExpansionLibrary

The API is not exactly the same, but they are fairly similar. I only had to make about 10 small changes to my sample apps - and everything still worked.


mattleibow:
approve

bholmes (commented):
We will have to add a ANDROID_HOME variable to the CI system, otherwise the gradle step will fail.  

Also as discussed on Slack you need to add a Readme.  Since there are no component docs the uses have no guidance on how to build the package or run the samples.  Some links to the Google docs will be good to have as well.

PR #177, opened 2017-07-13
Updated to Firebase Job Dispatcher 0.7.0.0

topgenorth:
New version of the Firebase Job Dispatcher on Github, this is the updated binding.

### DIFF
--- a/Android/FirebaseJobDispatcher/source/Firebase.JobDispatcher/Properties/AssemblyInfo.cs
+++ b/Android/FirebaseJobDispatcher/source/Firebase.JobDispatcher/Properties/AssemblyInfo.cs
@@ -9,5 +9,5 @@
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 
-[assembly: AssemblyVersion("0.7.0.0")]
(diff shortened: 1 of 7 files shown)


Android/FirebaseJobDispatcher/source/Firebase.JobDispatcher/Properties/AssemblyInfo.cs
```diff
@@ -9,5 +9,5 @@
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 
-[assembly: AssemblyVersion("0.6.0.0")]
-[assembly: AssemblyFileVersion("0.6.0.0")]
\ No newline at end of file
+[assembly: AssemblyVersion("0.7.0.0")]
```
bholmes: We do not  rev AssemblyVersion.  For bindings, the assembly version is usually just 1.0.0.0, and then we set the file version to the library version.  I suggest picking a version number and sticking with it. AssemblyFileVersion you can change.

bholmes (changes requested):
One change needed otherwise I think this is good to merge.

topgenorth:
Changed the AssemblyVersion.

PR #185, opened 2017-07-25
[PhotoView] Update to v1.3.1

mattleibow:
- added .nuspec
 - added external dependencies


mattleibow:
Updating this to 1.3.1 (the last of the v1.x series). This is a smaller set of changes and bugfixes. The v2.x series has some major changes with the package names and versioning. Some breaking changes are also going to happen.

PR #198, opened 2017-07-28
Add Firebase Push Notifications Plugin

rdelrosario:


jamesmontemagno:
For push notifications to be fully cross platform we are pushing towards Mobile Center. I will re-analyze though over time.

PR #200, opened 2017-07-31
[Kotlin] initial commit of binding

mattleibow:
Added a binding for the Kotlin Standard Library. It does not expose any types, just carries the native library. I don't think any types are available - Android Studio reports them as internal, but this may not be correct for all cases.

This is required for all libraries (.jar, .aar) that are built using the Kotlin language and compiler.

I am starting on 1.1.0, but 1.1.3-2 is out - trying to make sure we start far enough back to cover all bases (without starting too far back).

## Todo

**Libraries**
- [X] stdlib
- [x] stdlib-jre7
- [x] stdlib-jr8

**Sample**
- [X] native Kotlin library
- [X] sample app using Kotlin library

**Packaging**
- [X] NuGets for stdlib
- [x] NuGets for stdlib-jre7
- [x] NuGets for stdlib-jre8

**Legal**
- [x] License (https://github.com/JetBrains/kotlin/tree/master/license)


mattleibow:
@bholmes I may need advice on if or which of these third party licenses we need to add to our notes: https://github.com/JetBrains/kotlin/tree/master/license

mattleibow:
We also need to investigate whether we can/should do a single package with all the dependencies, or if they must be separate NuGets for each component.

Android/Kotlin/source/Xamarin.Kotlin.StdLib/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+[assembly: AssemblyProduct("")]
+[assembly: AssemblyCopyright("Copyright (c) 2017 Xamarin Inc")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// The assembly version has the format "{Major}.{Minor}.{Build}.{Revision}".
+// The form "{Major}.{Minor}.*" will automatically update the build and revision,
+// and "{Major}.{Minor}.{Build}.*" will update just the revision.
+
+[assembly: AssemblyVersion("1.1.0.0")]
```
Redth: This might cause issues for us between versions, we should maybe just use the standard 1.0.0.0 constant version number trick here?

bholmes:
@mattleibow as we discussed for licensing you will need in your Third Party Notices...

 - https://github.com/JetBrains/kotlin/blob/master/license/LICENSE.txt
 - https://github.com/JetBrains/kotlin/blob/master/license/NOTICE.txt
 - https://github.com/JetBrains/kotlin/blob/master/license/third_party/gwt_license.txt

Android/Kotlin/source/Xamarin.Kotlin.StdLib/Properties/AssemblyInfo.cs
mattleibow: Noted. Changed.

bholmes (approved):
These changes are good for me.

Redth (approved):

PR #232, opened 2017-09-18
[xbd] Add `ResolveAssemblyReferences` dependency

Redth:
It’s possible to invoke a build of a project via msbuild which does not hit a path causing `ResolveAssemblyReferences` target to run.

Eg:
```
msbuild /t:UpdateAndroidResources /p:DesignTimeBuild=true /p:BuildingInsideVisualStudio=true /v:diag AndroidApp.csproj
```

This adds the `ResolveAssemblyReferences` target as a dependency of both the iOS and Android _master_ targets in Xamarin.Build.Download to ensure they are run and our references are resolved.


bholmes (commented):
@Redth 

We forgot to discuss this in our one on one.  It makes sense.  We have always assumed that is complete at this point.  However if we do not specify it, it can not be guaranteed.  

I assume that you want to bump the version and publish?  Do you want to bump as part of this PR?  Let me know.

Redth:
@bholmes this is less critical now.  The bug was found in xamarin.android's master but apparently there's another msbuild property to use the old behaviour that will ultimately be called from the designer, circumventing this issue.

I think we still should add this to be explicit about the dependency, but it shouldn't affect users without.

bholmes:
OK.  So we will add for future a release.

PR #238, opened 2017-09-29
[Gigya] Update Android to 3.3.9

bholmes:


dalexsoto (approved):
LGTM 👍

rventuri76:
I tested all those changes and I can confirm that now we can successfully use the latest version of Gigya SDK both on android and iOS and for the first time we were able to make the GS bridge work correctly with out web view. thanks

mattleibow (approved):

PR #239, opened 2017-10-03
Update README

SKLn-Rad:
Please add my webview plugin to the list. Thanks!


jamesmontemagno:
We don't add UI components to plugins currently.

PR #241, opened 2017-10-10
Update NuGet dependencies, target Android 8.0, bind the newest .AAR

topgenorth:
Seems that it's possible to build this binding again in latest version of Visual Studio for Mac. Bugzilla [issue 59046](https://bugzilla.xamarin.com/show_bug.cgi?id=59046) was preventing this project from compiling. Whatever the issue was, it seems to be gone now.

Updated the:

* Target Android SDK (Android 8.0)
* NuGet dependencies
* Binds 0.8.3 of the Firebase Job Dispatcher .AAR

Call things a beta for now until I can see if any of the meta-data needs to be update as well.

### DIFF
--- a/Android/FirebaseJobDispatcher/Xamarin.Firebase.JobDispatcher.sln
+++ b/Android/FirebaseJobDispatcher/Xamarin.Firebase.JobDispatcher.sln
@@ -11,6 +11,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 	ProjectSection(SolutionItems) = preProject
 		build.cake = build.cake
 		Readme.md = Readme.md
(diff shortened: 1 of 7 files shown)


Redth:
Are you considering updating the nuget dependency in the nuspec as well?  I tend to keep the nuspec version matching what i compile against.

Also if you do that, might I suggest updating the nuspec's dependencies to declare a dependency on v4, but depend on all the things v4 depends on directly instead?

eg: 
```
<dependency id="Xamarin.Android.Support.Compat" version="25.4.0.2" />
<dependency id="Xamarin.Android.Support.Core.UI" version="25.4.0.2" />
<dependency id="Xamarin.Android.Support.Core.Utils" version="25.4.0.2" />
<dependency id="Xamarin.Android.Support.Fragment" version="25.4.0.2" />
<dependency id="Xamarin.Android.Support.Media.Compat" version="25.4.0.2" />
```

Redth:
LGTM

PR #242, opened 2017-10-13
[Estimote] Update the Android binding to 1.0.3

bholmes:
- Fixing the sample so that it compiles
 - Mostly namespace changes with a few class name changes
 - Removing the Permissions file from the binding project
   as I believe that we now merge the AndroidManifest from the
   aar file


bholmes:
- [x] I need to update the icon in the nuspec
- [x] We audit TP license and packaging
- [x] Review component docs

Issue #244, opened 2017-10-17
SlackTextViewController broke with iOS 11, needs emergency patch


softwaremills:
The iOS component SlackTextViewController broke in iOS 11. The devs checked in an emergency patch to slackhq/SlackTextViewController, but they've been working on a bunch of other stuff getting it ready for an official release. Meanwhile, it's been a month since iOS 11 dropped, with no official release in sight, and the devs said basically "look, just set your Podfile to source from the checkin instead of the latest version, and you'll be fine."

I've been trying to get Objective Sharpie to work for me, but overall, I'm just finding it difficult getting this thing working. The current XamarinComponents project has a bunch of .a files for different ABIs, but the Xcode project spits out .framework files, and I've been trying to get stuff working, but I'm no expert here.

Anyway, this means a key feature of our iOS app is dead in the water. I hate to complain about something free, but help? Please?

softwaremills:
Thread of the pull request with comments from people testing:
https://github.com/slackhq/SlackTextViewController/pull/624

Main dev comment: "If you are using Cocoapods or Carthage, you can checkout to a specific commit or branch. You don’t need a release to be able to update anyway."
https://github.com/slackhq/SlackTextViewController/issues/607#issuecomment-333606448

newky2k:
This has now been fixed with #246 and has been pushed and published

PR #246, opened 2017-10-19
[SlackTextViewController] Updated to (almost) v1.9.6 to fix iOS 11 bugs

mattleibow:
This addresses https://github.com/xamarin/XamarinComponents/issues/244

This build works off a specific commit due to the version not being out yet, and this is a critical fix.

### DIFF
--- a/iOS/SlackTextViewController/build.cake
+++ b/iOS/SlackTextViewController/build.cake
@@ -7,7 +7,7 @@ var IOS_PODS = new List<string> {
 	"platform :ios, '7.0'",
 	"install! 'cocoapods', :integrate_targets => false",
 	"target 'Xamarin' do",
-	"pod 'SlackTextViewController', '1.9.5'",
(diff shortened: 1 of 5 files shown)


softwaremills:
Thank you very much, sir!

mattleibow:
Because the native version is not yet released, this will be a preview release. As soon as the official version goes out, we will release the new version with any additional fixes/features.

softwaremills:
That seems eminently reasonable.

bholmes:
fyi...   https://www.nuget.org/packages/Xamarin.SlackTextViewController/1.9.6-preview

uhw76tgwquerhjbqwnejk:
does this fix the inserting/deleting cells issue?

softwaremills:
I'm not sure if it fixes that, but I have tested it and it does fix the problem where on iOS 11 the spacing at the bottom was all wrong and the text field didn't register taps. Thank you very much @mattleibow and @bholmes !

mattleibow:
Great to hear that it fixes the iOS 11 issue. Any other issues need to be logged against the native library as this is just the binding.

PR #247, opened 2017-10-27
Deleted old Dropbox v1 api

newky2k:
Deleted old Dropbox v1 api from XPlat


newky2k:
Old PR not working anymore

PR #254, opened 2017-11-20
[AndoidSwipeLayout] Added a NuGet and licensing

mattleibow:


mattleibow:
I ran a diff on the API and there is no difference between the old and the new. This just creates a nuget and adds the legal docs to the packages.

Android/AndroidSwipeLayout/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
bholmes, line 1: @mattleibow you now have 2 license files. Remove here and one on components folder.  Remove the components folder one and fix the yaml to point here.

bholmes (changes requested):

Android/AndroidSwipeLayout/LICENSE.md
mattleibow: Done.

bholmes (approved):

PR #258, opened 2017-11-27
[Estimote iOS] Update binding to 4.26.3

bholmes:

PR #270, opened 2017-12-10
Update existing iOS EstimoteSDK binding to latest cocoapods version

michaeldimoudis:
- Update iOS Estimote SDK from 4.7.3 to 4.26.3
- Fixed existing methods and delegates that were broken in current binding version, ie: SecureBeaconManager
- Did not add new method and delegate bindings between 4.7.3 and 4.26.3. Thought I'd keep this a simple update.
- I tried to add new method and delegate bindings between 4.7.3 and 4.26.3, such as ESTMonitoringV2Manager which I really wanted, binding did compile successfully, however I could not figure out how to actually use this in a Xamarin iOS project, due to the extra delegate in ESTMonitoringV2Manager :(

So basically, this PR is just a maintenance release to update and fix existing methods with update cocoapod. I did test the fixed SecureBeaconManager binding on an app on my device with secure beacons and it works great. Maybe in the future I'll continue with the full 4.26.3 binding, which is compiling in another branch but couldn't test ESTMonitoringV2Manager.


michaeldimoudis:
After I submitted this PR I saw this https://github.com/xamarin/XamarinComponents/pull/258, tested it with MonitoringV2Manager (I had a brain faze how to get it working in my app) and it works. Looks great. Thanks @bholmes! I'll close this now.

PR #284, opened 2018-01-08
Migrate JZMultiChoicesCircleButton to Nuget

newky2k:
Migrate JZMultiChoicesCircleButton to Nuget


iOS/JZMultiChoicesCircleButton/component/component.yaml
```diff
@@ -15,8 +15,12 @@ screenshots:
    - screenshots/00.jpg
 libraries:
   ios-unified: 
-    - ../source/JZMultiChoice/bin/release/JZMultiChoice.dll
+    - ../output/JZMultiChoice.dll
 samples:
```
bholmes, line 19: Sample needs to use Nuget

PR #288, opened 2018-01-12
[SWTableViewCell] to Nuget

adlair:


bholmes:
This needs to be redone.  It is mixed with other commits.

PR #300, opened 2018-02-06
[xbd] emit proguard configs to FileWrites always

Redth:
There was an issue with the proguard task where config files extracted and cached into the intermediate output dir were not being emitted to the FileWrites item group on subsequent incremental builds because the .stamp file was checked and then the task skipped if it existed. This PR now makes the cached proguard config file paths saved to the .stamp file so subsequent incremental builds can read them from the .stamp file and emit them to the FileWrites item group.


dellis1972:
Issue was reported at https://github.com/xamarin/xamarin-android/issues/1193

dellis1972 (approved):

dellis1972:
@Redth this looks good to me, you can probably merge it. I would do it myself but thats bad form ;)

Redth:
Yep, was waiting to roll this into another possible issue and release together.

PR #301, opened 2018-02-06
[xbd] Prevent possible race conditions accessing the assemblies for resource merging

Redth:
First big change is loading up and caching the `AssemblyName.GetAssemblyName(..)` calls in each execution of the task.  It should happen now in the `BuildRestoreMap` call once per task execution, and will also try to cache the results with the global build engine for parallel executions to utilize.  It also locks around the reading of the assembly name from file, as well as trying multiple times before failing.

The other big change is surrounding the access to the assembly for merging resources in an exclusive file lock, with some global build engine state gating too.

First we check to see if any other task instance has processed the assembly already, and then we try to open an exclusive lock on the .lock file for that output assembly.  If another task is already processing we will wait for a lock, and once we get the lock (the other task finished) we check the global build cache once more as it’s likely the other task already did what we were trying to do.

### DIFF
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/AsyncTask.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/AsyncTask.cs
@@ -80,7 +80,7 @@ namespace Xamarin.Build.Download
 				LogDebugMessage ("    {0}", item.ItemSpec);
 		}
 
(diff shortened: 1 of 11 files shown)


Redth:
[XbdRaceRepro.zip](https://github.com/xamarin/XamarinComponents/files/1701043/XbdRaceRepro.zip)

Here's a sln to help report it

PR #305, opened 2018-02-09
[GPUImage] to NuGet

adlair:


newky2k (approved):

newky2k (changes requested):
There are no changes that add Nuget capability for Shimmer as you've not added a nuspec file or updated the cake.build script to include one in the build process

PR #306, opened 2018-02-13
[JSQLMessageViewController] to NuGet

adlair:


newky2k (commented):
The change to shimmer appears in #305

iOS/JSQMessagesViewController/build.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
newky2k, line 1: Is this file necessary as build.cake has not been modified to call it and uses the common.cake.

newky2k (changes requested):
This update doesn't provide a nuspec file and doesn't modify the build.cake file to reference it

PR #312, opened 2018-02-20
[Shimmer] to Nuget.

adlair:


iOS/Shimmer/build.cake
```diff
@@ -3,10 +3,19 @@
 
 var TARGET = Argument ("t", Argument ("target", "Default"));
 
+var IOS_PODS = new List<string> {
+	"platform :ios, '11.2'",
```
SotoiGhost: Does this pod support only the most recent iOS version?

iOS/Shimmer/source/Shimmer/Structs.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using ObjCRuntime;
+
+namespace Shimmer.iOS
+{
+    [Native]
+public enum FBShimmerDirection : long
```
newky2k: Formatting in wrong in this file

iOS/Shimmer/nuget/Xamarin.Shimmer.nuspec
```diff
@@ -0,0 +1,18 @@
...
+  <metadata>
+    <id>Xamarin.Shimmer</id>
+    <title>Shimmer for Xamarin.iOS</title>
+    <version>1.0.2</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Shimmer is an easy way to add a shimmering effect to any view in your app.</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <projectUrl>https://components.xamarin.com/view/shimmer</projectUrl>
```
newky2k: This should be a FWLink as detail here https://gist.github.com/bholmes/39bad5c2296fe445df9e5bbf9261ab15

iOS/Shimmer/nuget/Xamarin.Shimmer.nuspec
```diff
@@ -0,0 +1,18 @@
...
+    <licenseUrl>https://components.xamarin.com/license/shimmer</licenseUrl>
```
newky2k: This should be a FWLink as detail here https://gist.github.com/bholmes/39bad5c2296fe445df9e5bbf9261ab15

iOS/Shimmer/source/Shimmer/Shimmer.nuspec
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
```
newky2k: This is a duplicate of the nuspec file that is in the nugets folder

newky2k (changes requested):

iOS/Shimmer/component/component.yaml
```diff
@@ -1,4 +1,4 @@
-version: 1.0.0.2
+version: 1.0.2
```
newky2k, line 1: This file needs to have the is_shell set and to reference the nuget that is specified in the spec. 

This is an example -> https://github.com/newky2k/XamarinComponents/blob/dh_InAppSettingsKit/iOS/InAppSettingsKit/component/component.yaml#L22-L24

newky2k (changes requested):
You should also specify nobuild: true on the sample.

Also does this need a external dependencies file for the cocoapod

You also need to move the licence.md file to the root folder from the components folder

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
```diff
@@ -0,0 +1,70 @@
...
+using Foundation;
+using ObjCRuntime;
+using CoreGraphics;
+
+namespace Shimmer.iOS
+{
+    // @protocol FBShimmering <NSObject>
+    [Protocol, Model]
+    [BaseType(typeof(NSObject))]
+    interface FBShimmering
```
SotoiGhost: I recommend removing the `FB` prefix. Also, you can add the `IFBShimmering` due `FBShimmering` is a Protocol.

SotoiGhost: According to Objective-C headers, there are two classes missing: `FBShimmeringLayer` and `FBShimmeringView`

iOS/Shimmer/External-Dependency-Info.txt
```diff
@@ -0,0 +1,44 @@
...
+Do not translate or localize
+
+Xamarin Components for Shimmer incorporates third party 
+material from the projects listed below. The original copyright notice and 
+the license under which Microsoft received such third party material are set 
+forth below.  Microsoft reserves all other rights not expressly granted, 
+whether by implication, estoppel or otherwise.
+
+########################################
+# Shimmer - Lincense
```
newky2k: Spelling mistake, should be license

iOS/Shimmer/component/component.yaml
```diff
@@ -16,8 +16,12 @@ libraries:
...
+  - name: "Shimmer Sample"
     path: ../samples/ShimmerSample/ShimmerSample.sln
-    configuration: "Release|iPhone"
+    installNuGets:
+      - project: ShimmerSample
+        packages:
+          - Xamarin.Shimmer
+    configuration: "Release|iPhone"    
 
 monodoc: ../docs/
```
newky2k: still needs is_shell: true

newky2k (changes requested):
Still needs a couple of changes

iOS/Shimmer/component/component.yaml
```diff
@@ -16,8 +16,12 @@ libraries:
   ios-unified: ../output/unified/Shimmer.dll
 
 samples:
-  - name: "Storyboard Sample"
+  - name: "Shimmer Sample"
     path: ../samples/ShimmerSample/ShimmerSample.sln
-    configuration: "Release|iPhone"
+    installNuGets:
```
newky2k: You need to add the remove line to remove the source project, before installing the nuget

https://github.com/xamarin/XamarinComponents/pull/311/files#diff-a928b56fc399e0e478f1ed32a6fd5ab6R22

iOS/Shimmer/component/component.yaml
```diff
@@ -16,8 +16,12 @@ libraries:
...
+      - project: ShimmerSample
+        packages:
+          - Xamarin.Shimmer
+    configuration: "Release|iPhone"
```
newky2k: need to add local-nuget-repo: ../output

iOS/Shimmer/component/component.yaml
newky2k: You need to add the externals file to the yaml output.  

https://github.com/xamarin/XamarinComponents/pull/311/files#diff-a928b56fc399e0e478f1ed32a6fd5ab6R33

newky2k (changes requested):
Still needs changes

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        // @required @property (getter = shimmeringHighlightLength, assign, readwrite, nonatomic, setter = setShimmeringHighlightLength:) CGFloat shimmeringHighlightWidth;
+        [Abstract]
+        [Export("shimmeringHighlightWidth")]
+        nfloat ShimmeringHighlightWidth { [Bind("shimmeringHighlightLength")] get; [Bind("setShimmeringHighlightLength:")] set; }
+
+        // @required @property (assign, readwrite, nonatomic) FBShimmerDirection shimmeringDirection;
+        [Abstract]
+        [Export("shimmeringDirection", ArgumentSemantic.Assign)]
+        FBShimmerDirection ShimmeringDirection { get; set; }
```
newky2k: This is failing to compile with the following error -> CS0246: The type or namespace name `FBShimmerDirection' could not be found. Are you missing an assembly reference?

newky2k (changes requested):
Compilation issue

iOS/Shimmer/External-Dependency-Info.txt
```diff
@@ -0,0 +1,44 @@
...
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
bholmes, line 44: I have noticed that this is a Facebook repo.  They have a [PATENTS](https://github.com/facebook/Shimmer/blob/master/PATENTS) file that needs to be accounted for.  You can use the solution that we have for [FacebookPop](https://github.com/xamarin/XamarinComponents/blob/453461dd7870f11fc34cbd11cb14372204552d48/iOS/FacebookPop/External-Dependency-Info.txt)

iOS/Shimmer/source/Shimmer/Properties/AssemblyInfo.cs
```diff
@@ -1,36 +1,34 @@
...
-// You can specify all the values or you can default the Build and Revision Numbers 
-// by using the '*' as shown below:
-// [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
+// The assembly version has the format "{Major}.{Minor}.{Build}.{Revision}".
+// The form "{Major}.{Minor}.*" will automatically update the build and revision,
+// and "{Major}.{Minor}.{Build}.*" will update just the revision.
+
+[assembly: AssemblyVersion("1.0.*")]
```
bholmes: remove the * from the version.  1.0.0.0

bholmes (changes requested):
I have inspected this PR only visually.  I did not pull it down and do a test build.  @newky2k can you test that?  

Also I noticed that this is now mixing a port of a existing library and not also including the bound native library to account for the new features.  Is this the right strategy or should we port the new features?

@mattleibow @Redth

iOS/Shimmer/source/Shimmer/Properties/AssemblyInfo.cs
newky2k: still need to change the assembly version still

newky2k (changes requested):

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
```diff
@@ -0,0 +1,60 @@
...
+using Foundation;
+using ObjCRuntime;
+using CoreGraphics;
+
+namespace Shimmer.iOS
+{
+    // @protocol FBShimmering <NSObject>
+    [Protocol, Model]
+    [BaseType(typeof(NSObject))]
+    interface Shimmering
```
SotoiGhost: Comparing with Objective-C header, there are some missing properties. Also, if you use this class as a Model, it will crash because it's missing the `Name = "FBShimmering"` to be able to change the protocol name.

SotoiGhost (changes requested):
I haven't downloaded the code and haven't tested it but, reviewing the code, I can see that there are some conflicts between the ported code and the bound code.

The new bound class is `Shimmering` and is flagged with `Protocol` and `Model` attributes. AFAIK, if you try to compile, it will try to create the `IShimmering` interface because of the `Protocol` attribute and the `IShimmering` already exists.

If the binding generator doesn't create the interface and only creates the Model class (just `Shimmering`), you will create a confusion to the user between the bound code and the ported code because, if you use the `IShimmering` interface, you will be using the ported code, and if you use the Model generated, you will be using the bound code, which is not a good idea.

Finally, supposing that the binding generator can successful create the interface, if you use it, which interface will it try to use… the generated one or the ported one?

I reviewed the Objective-C headers and just changed a little bit, I rather to create a binding over a port due is easier to maintain, but because this library is not update too often, you can continue porting the changed stuff.

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
bholmes: I agree with this opinion.  Let us change this over to be a 100% binding.

newky2k (changes requested):
The build still fails when run packaging(either the nuget or the component).  The error can be seen here -> https://gist.github.com/newky2k/e3f437ea989007a66209e6bb83061c72

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    // @protocol FBShimmering <NSObject>
+    [Protocol, Model]
+    [BaseType(typeof(NSObject), Name = "FBShimmering")]
+    interface Shimmering
+    {
+        // @required @property (getter = isShimmering, assign, readwrite, nonatomic) BOOL shimmering;
+        [Abstract]
+        [Export("shimmering")]
+        bool IsShimmering { [Bind("isShimmering")] get; set; }
```
SotoiGhost, line 17: This name is ok but I suggest changing this property to just `Shimmering` due it has a setter property. The `Is` prefix is recommended when is a read-only property. But I leave it to Adlair's decision.

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        // @required @property (assign, readwrite, nonatomic) CGFloat shimmeringSpeed;
+        [Abstract]
+        [Export("shimmeringSpeed")]
+        nfloat ShimmeringSpeed { get; set; }
+
+        // @required @property (getter = shimmeringHighlightLength, assign, readwrite, nonatomic, setter = setShimmeringHighlightLength:) CGFloat shimmeringHighlightWidth;
+        [Abstract]
+        [Export("shimmeringHighlightWidth")]
+        nfloat ShimmeringHighlightWidth { [Bind("shimmeringHighlightLength")] get; [Bind("setShimmeringHighlightLength:")] set; }
```
SotoiGhost, line 42: According to Obj-C headers, this property is now deprecated and replaced with `shimmeringHighlightLength` property.

iOS/Shimmer/source/Shimmer/ApiDefinition.cs
```diff
@@ -0,0 +1,86 @@
...
+    // @interface FBShimmeringView : UIView <FBShimmering>
+    [BaseType(typeof(UIView), Name = "FBShimmeringView")]
+    interface ShimmeringView : IShimmering
+    {
+        // @property (nonatomic, strong) UIView * contentView;
+        [Export("contentView", ArgumentSemantic.Strong)]
+        UIView ContentView { get; set; }
+    }
+
+    internal interface IShimmering : Shimmering
```
SotoiGhost, line 83: Is there a special reason to make this interface internal and make it implement the `Shimmering` interface?

iOS/Shimmer/source/Shimmer/Properties/AssemblyInfo.cs
```diff
@@ -1,36 +1,34 @@
...
-// You can specify all the values or you can default the Build and Revision Numbers 
-// by using the '*' as shown below:
-// [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
+// The assembly version has the format "{Major}.{Minor}.{Build}.{Revision}".
+// The form "{Major}.{Minor}.*" will automatically update the build and revision,
+// and "{Major}.{Minor}.{Build}.*" will update just the revision.
+
+[assembly: AssemblyVersion("1.0.2")]
```
SotoiGhost, line 28: @bholmes does this must be:

```
[assembly: AssemblyVersion("1.0.0.0")]
[assembly: AssemblyFileVersion("1.0.2.0")]
```

instead of the above stated?

bholmes: IIRC `AssemblyVersion` can not change.  `AssemblyFileVersion` and and should.  @mattleibow correct?

mattleibow: Yes.

newky2k (approved):

adlair:
I added changes on the PR https://github.com/xamarin/XamarinComponents/pull/366.

newky2k:
Reference #366.

PR #319, opened 2018-02-22
Android Wear

Redth:
Our binding source for Android Wear used to live in the PlayServicesComponents repository.  As google has shifted some content and libraries around for wear, it made more sense to move this out of the Play Services repo on its own.

This project now includes the bindings for:
 - `com.google.android.support.wearable`
 - `com.google.android.wearable.wearable`

There are other Wear related bindings that live elsewhere... These ones are still part of their respective bigger repositories (maven groups) so it makes sense to keep them there:
 - `com.google.android.gms.play-services-wearable` [xamarin/GooglePlayServicesComponents](https://github.com/xamarin/GooglePlayServicesComponents/tree/master/wearable)
 - `com.android.support.wear` [xamarin/AndroidSupportComponents](https://github.com/xamarin/AndroidSupportComponents/tree/master/wear)


Android/Wear/External-Dependency-Info.txt
```diff
@@ -0,0 +1,29 @@
...
+Xamarin Components for Google's Android Wear Libraries incorporates 
+third party material from the projects listed below. The original copyright 
+notice and the license under which Microsoft received such third party 
+material are set forth below.  Microsoft reserves all other rights not 
+expressly granted, whether by implication, estoppel or otherwise.
+
+########################################
+# Google Android Wear Libraries
+# https://developer.android.com/wear
+# https://dl.google.com/dl/android/maven2/com/google/android/support/wearable/2.2.0/wearable-2.2.0.pom
```
mattleibow: I think it may be better to split this into two separate nodes, just in case.

Android/Wear/nuget/Xamarin.Android.Wear.nuspec
```diff
@@ -0,0 +1,33 @@
...
+        <dependency id="Xamarin.Android.Support.Compat" version="26.1.0.1" />
+        <dependency id="Xamarin.Android.Support.Core.UI" version="26.1.0.1" />
+        <dependency id="Xamarin.Android.Support.Percent" version="26.1.0.1" />
+        <dependency id="Xamarin.Android.Support.v7.RecyclerView" version="26.1.0.1" />
+      </group>
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="nuget/Xamarin.Android.Wear.targets" target="build/MonoAndroid80" />
+    <file src="externals/support_wearable/proguard.txt" target="proguard/proguard.txt" />
```
mattleibow, line 29: Is there a specific reason for this to be in the root and not in the build directory? It may be better to not put too much in the root just in case they do something to nuget to either make this mean something, or do some validation with where stuff lives.

mattleibow (changes requested):
Looks good so far, just two comments.

Android/Wear/External-Dependency-Info.txt
Redth: You mean list the license once for each library instead?  Sure, not a bad idea.

Android/Wear/nuget/Xamarin.Android.Wear.nuspec
Redth: Perhaps that's where it should go instead.  That way we can stick MonoAndroid80 in front of it too...

mattleibow (approved):

PR #321, opened 2018-02-22
[constraint-layout] moved out of support repo

Redth:
This Android Support Constraint Layout and Solver set of bindings used to reside in xamarin/AndroidSupportComponents however it’s released independently of that set of libraries and makes more sense to be moved out of the repository.

This also updates the bindings to 1.1.0-beta5


Android/ConstraintLayout/samples/.vs/ConstraintLayoutSample/xs/UserPrefs.xml
```diff
@@ -0,0 +1,10 @@
+﻿<Properties StartupConfiguration="{17133E29-65C3-4B55-AC7E-107F50EE21D6}|Default">
```
mattleibow, line 1: this file should be removed

mattleibow: In my PR I added a .vs/ to .gitignore, so you can do the same and see who gets merged first :)

mattleibow (approved):

PR #331, opened 2018-03-07
[xbd] Add support for new `AndroidAarLibrary` item

Redth:
Xamarin.Android in 15.7 and newer will include the `AndroidAarLibrary` item group.  .aar files added to this group will automatically be compiled into the app, so Xamarin.Build.Download no longer has to embed them into the referenced assembly that needs them as embedded resources.

Current builds of 15.7 do not include the `<AvailableItemName Include="AndroidAarLibrary" />` so this PR manually adds it to our targets for now to be able to test with.  This will need to be removed before merging.


Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.targets
```diff
@@ -107,9 +107,23 @@
 		</ItemGroup>
 	</Target>
 
+	<!-- TODO: AndroidAarLibrary is not yet in AvailableItemName so we added it manually for now -->
+	<ItemGroup>
+		<AvailableItemName Include="AndroidAarLibrary" />
```
Redth: This is temporary until 15.7 builds include it.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.targets
```diff
@@ -107,9 +107,23 @@
...
+	</ItemGroup>
+	<!-- END TODO -->
+		
+	<Target Name="_XamarinBuildDownloadAarInclude"
+			DependsOnTargets="_XamarinBuildDownload"
+			Condition="'@(XamarinBuildDownloadRestoreAssemblyAar)'!='' and %(AvailableItemName.Identity) == 'AndroidAarLibrary'">
```
dellis1972: does `%(AvailableItemName.Identity)` need to be in single quotes?

dellis1972: Thinking about it , it might be better to define a property 

```xml
<PropertyGroup>
   <SupportsAarLibrary Condition="'%(AvailableItemName.Identity)' == 'AndroidAarLibrary'">true</SupportsAarLibrary>
<SupportsAarLibrary Condition="'$(SupportsAarLibrary)' == ''">false</SupportsAarLibrary>
</PropertyGroup>
```

I know MSbuild might run a target multiple times when using `%(foo.Identity)` due to its Batching support. I'm not sure if that happens in this case. It might not, if it doesn't ignore me :)

dellis1972 (changes requested):

dellis1972 (approved):

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.targets
```diff
@@ -107,9 +119,17 @@
 		</ItemGroup>
 	</Target>
 
+	<Target Name="_XamarinBuildDownloadAarInclude"
+			DependsOnTargets="_XamarinBuildDownload"
+			Condition="'@(XamarinBuildDownloadRestoreAssemblyAar)'!='' and '$(SupportsAarLibrary)'' == 'true'">
```
mattleibow: There is an extra `'` here: `'$(SupportsAarLibrary)**''** == 'true'`

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.targets
```diff
@@ -107,9 +119,17 @@
...
+		<ItemGroup>
+			<AndroidAarLibrary Include="@(XamarinBuildDownloadRestoreAssemblyAar)" />
+		</ItemGroup>
+	</Target>
+
 	<Target Name="_XamarinAndroidBuildAarRestore"
 			DependsOnTargets="_XamarinBuildDownload"
-			Condition="'@(XamarinBuildDownloadRestoreAssemblyAar)'!=''"
+			Condition="'@(XamarinBuildDownloadRestoreAssemblyAar)'!='' and '$(SupportsAarLibrary)'' != 'true'"
```
mattleibow: Another double `''` here - that autocomplete :0

mattleibow (changes requested):

Redth:
Added to #696

Issue #84, opened 2016-11-09
MapBox SDK iOS


Julien-Mialon:
Hello,

Great job on the MapBox ios SDK, I just have one issue with this, when trying to use offline storage, I can't access fields of struct OfflinePackProgress (https://github.com/xamarin/XamarinComponents/blob/master/XPlat/Mapbox/iOS/source/MapboxSDK.iOS/StructsAndEnums.cs#L73-L78)

I can guess it's because they are not marked as public, but is there any reason for this ?

Thanks,
Julien.

Redth:
We have deprecated our bindings for Mapbox and have delisted the existing NuGet packages for them.

Please see the [README](https://github.com/xamarin/XamarinComponents/blob/master/XPlat/Mapbox/README.md) for more information.

PR #345, opened 2018-04-09
Make the properties of the OfflinePackProgress struct public so they …

WiredUK:
…can be used


WiredUK:
Fixing [issue 84](https://github.com/xamarin/XamarinComponents/issues/84)

newky2k (changes requested):
Android source, at least is not compiling ->
https://gist.github.com/newky2k/7ab69a22a182f6c67382d5497e4313f6

WiredUK:
@newky2k I'm confused about a couple of things on this one. First of all the fact it doesn't compile when all I did was make some struct properties public. Secondly the gist you added states that Android is not compiling when the change I made was for the iOS code only. Colour me baffled! :)

newky2k:
@WiredUK  Hi,  I'm not suggesting your code is the cause of the failure just that overall, in its current state the component doesn't compile, and someone from the xamarin team will need to investigate before it can be merged

PR #348, opened 2018-04-10
[AnimatedCheckButton] to Nuget.

adlair:


iOS/AnimatedCheckButton/External-Dependency-Info.txt
```diff
@@ -0,0 +1,31 @@
+THIRD-PARTY SOFTWARE NOTICES AND INFORMATION
```
mattleibow, line 1: We shouldn't need this file since I mostly wrote this code.

mattleibow: @bholmes Is this assumption right?

mattleibow (dismissed):

iOS/AnimatedCheckButton/External-Dependency-Info.txt
bholmes: @mattleibow well did you mostly write it or did you write it?  If you actually rewrote the entire library I agree we do not need it.  However if you did draw some inspiration from the code I feel it is fair to give credit by providing attribution.

adlair:
I added new changes on https://github.com/xamarin/XamarinComponents/pull/376.

newky2k:
Replace by #376

PR #349, opened 2018-04-10
[MBAlertView] to Nuget.

adlair:


newky2k (changes requested):
Needs to have iphone x support

newky2k (approved):

PR #357, opened 2018-04-20
[JASidePanels] iPhone X support added.

adlair:


newky2k (changes requested):
JASidePanelsSample doesn't have a launchscreen and the other samples are not set to have the launchscreen used so they don't show full screen on iphone x

newky2k (approved):

PR #369, opened 2018-05-02
[xbd] Fix errors with partial zip downloads

Redth:
Previously we were requesting multiple ranges in a single request, which now seems unreliable.  This changes to using a single http request for each part, while still specifying a requested range for only the part of the zip file needed.

Also updated nuget dependency


Util/Xamarin.Build.Download/source/Xamarin.Build.Download/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.AspNet.WebApi.Client" version="5.2.3" targetFramework="net45" />
+  <package id="Microsoft.AspNet.WebApi.Client" version="5.2.4" targetFramework="net45" />
```
bholmes, line 3: Did you intend to update this?

Redth:
Yes I did.

PR #376, opened 2018-05-11
[AnimatedCheckButton] to Nuget.

adlair:


newky2k (changes requested):
This needs to be updated to support iPhone X in the sample.  this is done by adding a launchscreen file and setting in the info.plist

newky2k (approved):

PR #388, opened 2018-05-23
[AnimatedShapeButton] to Nuget.

adlair:


iOS/AnimatedShapeButton/source/AnimatedShapeButton/AnimatedShapeButton.designer.cs
```diff
@@ -0,0 +1,20 @@
...
+// actions declared in your storyboard file.
+// Manual changes to this file will not be maintained.
+//
+using Foundation;
+using System;
+using System.CodeDom.Compiler;
+
+namespace AnimatedButtons
+{
+    [Register ("AnimatedShapeButton")]
```
mattleibow: I think this file is not needed - it gets generated incorrectly.

mattleibow (dismissed):

iOS/AnimatedShapeButton/source/AnimatedShapeButton/AnimatedShapeButton.designer.cs
adlair: Thanks @mattleibow, I removed it.

PR #398, opened 2018-05-30
[ARCore] Update to released 1.2.0

vicfergar:
- Bump the aar download version


vicfergar:
Not necessary since #414

PR #404, opened 2018-06-08
[JSQMessagesViewController] Updated to 7.3.5.

adlair:


iOS/JSQMessagesViewController/build.cake
```diff
@@ -7,7 +7,7 @@ var IOS_PODS = new List<string> {
 	"platform :ios, '7.0'",
 	"install! 'cocoapods', :integrate_targets => false",
 	"target 'Xamarin' do",
-	"pod 'JSQMessagesViewController'",
+	"pod 'JSQMessagesViewController', '~> 7.3'",
```
Redth: We should use an explicit version here `= 7.3.5`

Redth (changes requested):
Please fix up the cocoa pods version.

Redth (approved):

PR #407, opened 2018-06-14
[Task631160] AndroidCodegenTarget added on Android projects.

adlair:


Redth (changes requested):
This is a bit picky of a request, but the property name should be `AndroidCodegenTarget` instead of `AndroidCodeGenTarget`... I don't think that case really matters here, but it is an easy find/replace fix, and is more correct :)

Redth (approved):

PR #408, opened 2018-06-14
[Task629208] Add missing version checkers.

adlair:


Redth (changes requested):
@adlair Please update and add any missing version checkers for Google API's for iOS and Firebase iOS bindings as we discussed

PR #413, opened 2018-06-24
[iRate] 1.12.2 update

yuriy-blinov:
Update iRate binding to the latest version available to support SKStoreReviewController.


newky2k:
hi @yuriy-blinov,

The source project has now been deprecated and work has ceased so we will not be providing any further updates to the package.  

Thanks for you contributions

PR #422, opened 2018-07-17
[iRate] Updated to v1.12.2.

adlair:


bholmes:
I think that we should leave this PR out because iRate is no longer supported.

https://github.com/nicklockwood/iRate

@Redth - we should delist the NuGet

https://www.nuget.org/packages/Xamarin.iRate/

newky2k:
This project has been deprecated and will not receive any further updates

PR #428, opened 2018-07-27
[Crashlytics] Bindings done.

adlair:


jldubz:
Is there an ETA for when this will be approved?  There are less than 3 weeks remaining until current crash analytics will be turned off, and I would appreciate some time to test implementing the new libraries before then.

gerardmrk:
Also requesting for an ETA. We're currently needing to install a crash-reporter and would prefer using Crashlytics over other solutions out there. 

If we know the approximate date this gets published, we can decide if it's worth waiting for it or whether to just use another alternative like Microsoft's App Center.

Appreciate any info. Please note I am not trying to push for a faster release, I'm just requesting for an estimated date this will be completed by, so we can make an informed decision. Thanks!

whithajess:
The feature branch looks to be in the PR stage now: https://github.com/xamarin/XamarinComponents/pull/441

Redth:
Yes we are reviewing the PR and should have something released this week. I've done some basic testing and am able to get crashes showing up in firebase console.

whithajess:
Awesome - thanks @Redth

gerardmrk:
Thanks @Redth

PR #429, opened 2018-07-27
TensorFlow.Lite and Protobuf.Lite added

moljac:

PR #430, opened 2018-07-27
Update manifest.yaml

moljac:


moljac:
TensorFlow.Lite and Protobuf.Lite added 
Duplicate of #429

newky2k (changes requested):
@moljac  as the nuget package is defined in the sdk style csproj, then you need to modify the build.cake for TensorFlow and Protobuf to work more like this -> https://github.com/xamarin/XamarinComponents/blob/master/Android/AndroidThings/build.cake#L52-L69

Also when compiling Protobuf with 
"sh ../../build.sh -t=libs"
I got an error ->
https://gist.github.com/newky2k/0481599e6508e8c75759938cf365dad7

moljac:
OK. I will check that, but it builds locally on my Mac.

moljac:
Fixed cake script was downloading jar and in the cake was aar

I had to create new branch and PR for this one

https://github.com/xamarin/XamarinComponents/pull/431

PR #468, opened 2018-10-14
Update plugins links

MauroCoppola:
Hi, I updated link to GitHub repository for two plugins since one was broken and another one pointed to the deprecated version of the project.


newky2k (approved):
Both the links are valid and approved

PR #500, opened 2018-12-18
upgrade DZNEmptyDataSet to 1.8.1 and update sample code

Wenfengcheng:
1. upgrade DZNEmptyDataSet to 1.8.1 
2. update sample project


newky2k:
@Wenfengcheng  could you create an issue for this Pull request and like this PR to it.

That way we can link it into our DevOps

Regards

Wenfengcheng:
Fix #510

newky2k (changes requested):
@Wenfengcheng Thank you for your submission, however there are some changes that need to be made before we can accept it primarily the inclusion of the framework directly in the source code.

Using the externals build task, build.cake, you can use cocoapods to download the latest version of the library as part of the build process.  You can then update the binding and use the latest version.  This may be as simple as changing the version of the pod download here -> https://github.com/xamarin/XamarinComponents/blob/master/iOS/DZNEmptyDataSet/build.cake#L10

Additionally, once the component is updated to use the new version you need to update the nuspec file too.

If you have any questions please let me know.

iOS/DZNEmptyDataSet/source/DZNEmptyDataSet/DZNEmptyDataSet.csproj
```diff
@@ -46,15 +46,13 @@
...
-      <Link>libDZNEmptyDataSet.a</Link>
-    </ObjcBindingNativeLibrary>
+    <Compile Include="Additions.cs" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="libDZNEmptyDataSet.linkwith.cs">
-      <DependentUpon>libDZNEmptyDataSet.a</DependentUpon>
-    </Compile>
-    <Compile Include="Additions.cs" />
+    <NativeReference Include="DZNEmptyDataSet.framework">
```
newky2k: I don't think these changes are correct when using the .a file from the pod

newky2k (changes requested):
Hi,

The DZEmptyDataSet.csproj is still looking at the binary for the framework in the DZNEmptyDataSet folder, which shouldn't have been added.  The pod download loads and compiles the libDZNemptyDataset.a file so there is no need to have the framework folder in the PR.

iOS/DZNEmptyDataSet/source/DZNEmptyDataSet/DZNEmptyDataSet.csproj
Wenfengcheng: ok, I'll fix that.

newky2k (approved):

PR #504, opened 2019-01-18
[MaterialComponents] Added some overloads to help alleviate the type changes since v60

mattleibow:
Added some overloads to help alleviate the type changes since v60:
 - the type used for elevation was changed to `nfloat`, so I added back the original methods for `double`
 - some properties and getter methods are still "broken", but that is harder to fix as the return type changed

The most notable bug so far is this on: https://github.com/xamarin/Xamarin.Forms/issues/5021

### DIFF
--- a/iOS/MaterialComponents/source/MaterialComponents/Extensions.cs
+++ b/iOS/MaterialComponents/source/MaterialComponents/Extensions.cs
@@ -93,6 +93,30 @@ namespace MaterialComponents {
 		}
 	}
 
+	partial class Card {
+		public void SetShadowElevation (double shadowElevation, UIControlState state) {
+			SetShadowElevation ((nfloat)shadowElevation, state);
+		}
+	}
+
+	partial class CardCollectionCell {
+		public void SetShadowElevation (double shadowElevation, CardCellState state) {
+			SetShadowElevation ((nfloat)shadowElevation, state);
+		}
+	}
+
+	partial class ChipView {
+		public void SetElevation (double elevation, UIControlState state) {
+			SetElevation ((nfloat)elevation, state);
+		}
+	}
+
+	partial class Button {
+		public void SetElevation (double elevation, UIControlState state) {
+			SetElevation ((nfloat)elevation, state);
+		}
+	}
+
 	partial class TextFieldColorThemer {
 		public static void ApplySemanticColorSchemeToAll (IColorScheming colorScheme, Type textInputControllerType) {
 			ApplySemanticColorSchemeToAll (colorScheme, new Class (textInputControllerType));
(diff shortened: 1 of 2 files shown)


iOS/MaterialComponents/source/MaterialComponents/Extensions.cs
```diff
@@ -93,6 +93,30 @@ partial class BottomAppBarColorThemer {
 		}
 	}
 
+	partial class Card {
+		public void SetShadowElevation (double shadowElevation, UIControlState state) {
```
SotoiGhost: I would suggest adding the method with a `Wrap` attribute below the original signature

```csharp
[Wrap ("SetShadowElevation ((nfloat)shadowElevation, state)")]
void SetShadowElevation (double shadowElevation, UIControlState state);
```

But that's my feel. Either way, LGTM. 👍🏼

SotoiGhost (dismissed):

iOS/MaterialComponents/source/MaterialComponents/Extensions.cs
mattleibow: DONE!

iOS/MaterialComponents/source/MaterialComponents/Extensions.cs
```diff
@@ -80,80 +80,27 @@ public static class ShadowElevations {
...
-		public static void ApplyFontSchemeToTextField (IFontScheme fontScheme, TextField textField) {
-			ApplyFontScheme (fontScheme, textField);
-		}
 	}
 	
 	partial class TextFieldTypographyThemer {
-		public static void ApplyTypographySchemeToAll (ITypographyScheming typographyScheme, Type textInputControllerType) {
-			ApplyTypographySchemeToAll (typographyScheme, new Class (textInputControllerType));
-		}
 		public static void ApplyTypographySchemeToAll<T> (ITypographyScheming typographyScheme) where T : ITextInputController {
```
SotoiGhost, line 96: Can you do …?

```csharp
[Static]
[Wrap ("ApplyTypographySchemeToAll (typographyScheme, typeof (T))")]
void ApplyTypographySchemeToAll<T> (ITypographyScheming typographyScheme) where T : ITextInputController;
```

If doesn't work, everything else, LGTM!

SotoiGhost (approved):

iOS/MaterialComponents/source/MaterialComponents/Extensions.cs
mattleibow: Sadly no... It looses the generics information 😭

SotoiGhost: @dalexsoto Could this be a bug?

dalexsoto: Nope, Wrap is not intended to be used that way https://github.com/xamarin/xamarin-macios/issues/5444#issuecomment-456086890

PR #506, opened 2019-01-24
TensorFlow Lite (Android) update to 1.12.0

newky2k:
Updated to 1.12.0 and added nuget diff tools for comparison with previous version


XPlat/TensorFlow.Lite/source/Xamarin.TensorFlow.Lite.Bindings.XamarinAndroid/Xamarin.TensorFlow.Lite.Bindings.XamarinAndroid.csproj
```diff
@@ -42,17 +42,17 @@
...
     </PropertyGroup>
 
     <ItemGroup>
         <TransformFile Include="transforms\*.xml" />
     </ItemGroup>
 
     <ItemGroup>
-        <LibraryProjectZip Include="..\..\externals\android\tensorflow-lite-1.9.0.aar">
-            <Link>Jars\tensorflow-lite-1.9.0.aar</Link>
+        <LibraryProjectZip Include="..\..\externals\android\tensorflow-lite-1.12.0.aar">
```
Redth: What if we download this to just a `tensorflow-lite.aar` file and leave out the version so we don't have to bump it each time?

XPlat/TensorFlow.Lite/build.cake
```diff
@@ -132,6 +146,47 @@ Task ("clean")
 		}
 	);
 
+Task ("docs-api-diff")
+	.IsDependentOn ("nuget")
+    .Does (async () =>
+{
+    var baseDir = "./output/api-diff";
```
Redth: I wonder if we should refactor this out to common.cake as well and just make it a simple method call from here with the bare minimum parameters passed in...

XPlat/TensorFlow.Lite/source/Xamarin.TensorFlow.Lite.Bindings.XamarinAndroid/Xamarin.TensorFlow.Lite.Bindings.XamarinAndroid.csproj
newky2k: changed

XPlat/TensorFlow.Lite/build.cake
newky2k: moved to common.cake and added BuildApiDiff method

XPlat/TensorFlow.Lite/build.cake
```diff
@@ -45,16 +45,30 @@ Running Cake to Build targets
 		mono tools/Cake/Cake.exe --verbosity=diagnostic --target=nuget
 #########################################################################################
 */
+#load "../../common.cake"
+//#load "apidiffutils.cake"
```
mattleibow: We probably can delete this line...

XPlat/TensorFlow.Lite/build.cake
```diff
@@ -132,6 +146,13 @@ Task ("clean")
 		}
 	);
 
+Task ("docs-api-diff")
+	.IsDependentOn ("nuget")
+    .Does (async () =>
+{
+	await BuildApiDiff(NUGET_PACKAGE_ID, NUGET_VERSION);
```
mattleibow: We may just want to run this after the "nuget" task in the base. For all output nugets.

build.cake
```diff
@@ -2,6 +2,11 @@
 
 #addin nuget:?package=redth.xunit.resultwriter&version=1.0.0
 #addin "nuget:?package=Xamarin.Nuget.Validator&version=1.1.1"
+#addin nuget:?package=Mono.ApiTools.NuGetDiff&version=1.0.1&loaddependencies=true
+
+using Mono.ApiTools;
+using NuGet.Packaging;
+using NuGet.Versioning;
```
mattleibow: do we need this here? most of the logic moved to common.cake

common.cake
```diff
@@ -319,3 +323,51 @@ void DownloadMonoSources (string tag, DirectoryPath dest, params string[] urls)
 		}
 	}
 }
+
+/// Api Diff Stuff
+
+async Task BuildApiDiff (string packageId, string currentVersionNo)
```
mattleibow: I deleted all the SkiaSharp-specific code because I de things to make it nicer for committing back to the repo. And, i have both a specific package structure, and ID style. That logic would have caused issues for most packages that had more than one platform (namely, dlls would be ignored))

common.cake
```diff
@@ -319,3 +323,51 @@ void DownloadMonoSources (string tag, DirectoryPath dest, params string[] urls)
...
+
+	// create comparer
+	var comparer = new NuGetDiff ();
+	comparer.PackageCache = "./externals/package_cache";
+	comparer.SaveAssemblyApiInfo = true;       // we don't keep this, but it lets us know if there were no changes
+	comparer.SaveAssemblyMarkdownDiff = true;  // we want markdown
+	comparer.IgnoreResolutionErrors = true;    // we don't care if frameowrk/platform types can't be found
+
+	// compare
+	using (var reader = new PackageArchiveReader ($"./output/{packageId.ToLower ()}.{currentVersionNo}.nupkg"))
```
mattleibow: This may be dodgy for some packages, so it may actually be better to just pass the nuspkg path to the method, and then open the reader. The reader will then give you the id and version, which you can use to pull down the latest version, and then run the compare:

```csharp
async Task BuildApiDiff (FilePath nupkg) {
    using (var reader = new PackageArchiveReader (nupkg.FullPath))  {
        var packageId = reader.GetIdentity ().Id;
        // do everything ...
    }
}
```

We can even go further and make an overload that takes a folder or a collection of files so we can diff whole chunks.

newky2k: i've remapped it to work like this and moved it to the root build.cake file

newky2k:
Fixes #509

build.cake
```diff
@@ -380,4 +380,19 @@ Task("nuget-validation")
 	}
 
 });
+
+Task ("docs-api-diff")
```
mattleibow, line 384: What starts this task? Will it always run? Or is it manual?

newky2k: it'll need to be added into ci the same as the nuget-validation task

Redth (approved):

PR #517, opened 2019-01-31
Updating Google VR to v1.180

### LINKED ISSUE #129: Google VR update library and project samples about Android and IOs VideoWidgetView
Hello everyone. I am building for a company a cross platform application Xamarin environment and to be able to do what I was required to use GoogleVR component, especcialy the videowidget tool, for iOS and Android inside XPlat folder. My problem is reflected in outdated versions of Google VR libraries (the last was 0.8.0, now is 1.3.0) to use and therefore can not apply newer methods and also there is nothing about an explanation to import correctly the library for his metadata and transform reference. Someone managed to operate the native library binding through the reference solutions for the new library references with a sample code? If so I would be really grateful needing a proper tutorial.

> VS bug [#732190](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/732190)

newky2k:
fixes #129

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ externals
 *.userprefs
 .vs/
 tmpnupkg/
-
+.user
 .DS_Store
 XPlat/DropboxApiV2/samples/DropboxV2ApiSampleDroid/.vs/DropboxV2ApiSampleDroid/xs/UserPrefs.xml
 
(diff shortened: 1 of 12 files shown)


newky2k:
gettting build issues -> https://gist.github.com/newky2k/b18b5023b1655728dd213745720681d8

KRA2008:
i'd love to see this happen or hear about some path forward. from my research it looks like the path to using Cardboard in Xamarin has been lost. the Cardboard SDK was even open sourced since this issue was last updated.

PR #532, opened 2019-02-22
[Update] Kotlin v1.3.21

SotoiGhost:
`Xamarin.Kotlin.StdLib.Jre*` was replaced with `Xamarin.Kotlin.StdLib.Jdk*`


Android/Kotlin/nuget/Xamarin.Kotlin.StdLib.Jdk7.nuspec
```diff
@@ -1,22 +1,22 @@
 <?xml version="1.0"?>
 <package >
   <metadata>
-    <id>Xamarin.Kotlin.StdLib.Jre7</id>
-    <title>Kotlin Standard Library JRE 7 extension for Xamarin.Android</title>
-    <version>1.3.11.0</version>
+    <id>Xamarin.Kotlin.StdLib.Jdk7</id>
```
mattleibow, line 4: Did they just up and rename the id? JRE is dead, JDK is the new JRE?

mattleibow (approved):

Android/Kotlin/nuget/Xamarin.Kotlin.StdLib.Jdk7.nuspec
SotoiGhost: Yep, according to this official [doc][1]. They state that:

> The Kotlin standard library is now fully compatible with the Java 9 module system, which forbids split packages (multiple jar files declaring classes in the same package). In order to support that, new artifacts kotlin-stdlib-jdk7 and kotlin-stdlib-jdk8 are introduced, which replace the old kotlin-stdlib-jre7 and kotlin-stdlib-jre8.

[1]: http://kotlinlang.org/docs/reference/whatsnew12.html#kotlin-standard-library-artifacts-and-split-packages

PR #536, opened 2019-03-12
Update Guava to 27.1

Redth:


Android/Guava/build.cake
```diff
@@ -1,28 +1,27 @@
...
-#load "../../common.cake"
-
 var TARGET = Argument ("t", Argument ("target", "Default"));
 
-var GUAVA_VERSION = "23.2-android";
-var NUGET_VERSION = "23.2.0";
+var GUAVA_NUGET_VERSION = "27.1";
+var GUAVA_FAILUREACCESS_NUGET_VERSION = "1.0.1";
+
+var GUAVA_VERSION = GUAVA_NUGET_VERSION + "-android";
```
SotoiGhost, line 6: Just as a note: cake now supports [String Interpolation][1], if you like it more than concatenation and `string.Format`.

[1]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

Android/Guava/source/Guava/Guava.csproj
```diff
@@ -1,64 +1,81 @@
...
+    </None>
+    <None Include="..\..\External-Dependency-Info.txt" Pack="True" PackagePath="THIRD-PARTY-NOTICES.txt" />
   </ItemGroup>
+
   <ItemGroup>
-    <TransformFile Include="Transforms\EnumFields.xml" />
-    <TransformFile Include="Transforms\EnumMethods.xml" />
-    <TransformFile Include="Transforms\Metadata.xml" />
+    <EmbeddedReferenceJar Include="..\..\externals\guava.jar" />
+    <!-- <ReferenceJar Include="..\..\externals\animal-sniffer-annotations.jar" /> -->
```
SotoiGhost: Should these comments be removed?

Redth: Yeah they should be for good measure, though they aren't really needed since we're now just using EmbeddedReferenceJar and aren't producing any API bindings.  It's good to keep them uncommented for future releases if we do change this though, thanks!.

SotoiGhost (approved):
LGTM! 👍🏼

PR #537, opened 2019-03-13
[Guava] Added ListenableFuture binding and made dependencies sane again

Redth:
This adds a binding to ListenableFuture and fixes up some weird dependency implementations on Google's side...

A bit of history:
- Firebase Firestore for Android depends on `com.google.guava:guava`
- WorkManager depends on `com.google.guava:listenablefuture:1.0`

But...
- `com.google.guava:guava` contains the interface `com.google.common.util.concurrent.ListenableFuture`
- `com.google.guava:listenablefuture` _also_ contains this interface

To fix this duplicate type problem in the event of referencing both....
- `guava` depends on >= `listenablefuture` version >= `9999.0-empty-to-avoid-conflict-with-guava`
- `listenablefuture:9999.0-empty-to-avoid-conflict-with-guava` is an empty package, it does not contain the interface `com.google.common.util.concurrent.ListenableFuture`

So, instead of making `listenablefuture` a normal dependency of `guava` and moving the `ListenableFuture` implementation out of `guava` and *only* into `listenablefuture`, they did that weird version trick.

This matters because:
- `guava` is ~2.3mb
- `listenablefuture` is ~3kb (it is literally a .jar file with one interface with one method)
- We don't want users of `WorkManager` to incur 2.3mb of overhead if they aren't using anything else which needs the full `guava` and they only need 3kb.

We could:
a. Do a weird version thing with our nuget packages like Google does
b. Do some .targets trickery to mess with `ReferencedAssemblies`
c. Remove `listenablefuture` from `guava.jar` and make our `guava` binding depend on our `listenablefuture` binding and the `ListenableFuture` implementation just always lives there

This is choosing option *c*.

### DIFF
--- a/Android/Guava/Guava.sln
+++ b/Android/Guava/Guava.sln
@@ -7,6 +7,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Guava", "source\Guava\Guava
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Guava.FailureAccess", "source\Guava.FailureAccess\Guava.FailureAccess.csproj", "{F77FD2C9-181F-43C9-9533-20EE845F74BC}"
 EndProject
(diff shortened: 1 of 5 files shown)


SotoiGhost (approved):
LGTM! 👍🏼

PR #547, opened 2019-03-28
Get the Mono.Posix build onto DevOps

mattleibow:
The first step is to just use make. The next will be to properly integrate it with the components builds.


mattleibow:
The build now succeeds: https://dev.azure.com/xamarin/public/_build?definitionId=24

newky2k (approved):

Issue #464, opened 2018-10-09
Update Glide to the latest


rubo:
Currently, the Xamarin.Android.Glide has v4.0.0 while Glide already reached to v4.8.0. Please consider updating, it's a must-have library.

> VS bug [#732178](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/732178)

newky2k:
@rubo we can look to add this in to the roadmap.  If you have the time you could provide an update as a pull request for us to review

adrianknight89:
That last 2017 update raises all the alarm bells for me. Would love to see this updated soon.

Redth:
Would love some community help here!

jamesmontemagno:
Have you seen: https://github.com/jonathanpeppers/glidex maybe we can bring this in officially with @jonathanpeppers ?

jonathanpeppers:
I would love to just drop my binding and use an official Xamarin one.

It's weird I swear I looked for this component, but then ended up binding glide myself.

jonathanpeppers:
I'm using 4.7.0: https://github.com/jonathanpeppers/glidex/blob/8363cade7de117c08b59b51bd63f825aa867e938/glidex/glidex.csproj#L54

And the only other improvement I have is `[assembly:LinkerSafe]`: https://github.com/jonathanpeppers/glidex/blob/8363cade7de117c08b59b51bd63f825aa867e938/glidex/Properties/AssemblyInfo.cs#L16

rubo:
@jonathanpeppers Why does glidex use v4.7.0? No time to update or technical difficulties because of the lack of a single JAR?

jonathanpeppers:
For some reason they quit releasing jar files here: https://github.com/bumptech/glide/releases

jonathanpeppers:
I presume a workaround would be to make a gradle project to download everything from maven and bundle into a single, self-contained jar.

I did that in xamarin-android for r8 here: https://github.com/xamarin/xamarin-android/blob/d7dcc5b5b9c2e2466e4d39769e3dbdf378613c91/src/r8/build.gradle

I've not had time to do this in glidex yet, but I don't know if the components guys here take a different approach.

xyfoo:
> For some reason they quit releasing jar files here: https://github.com/bumptech/glide/releases

I think even an update from v4.0.0 to v4.7.0 is still significant enough for most folks.

rubo:
> Would love some community help here!

[...]

PR #548, opened 2019-04-01
Upgraded to Glide 4.9.0

rubo:
This PR addresses the issue #464.
As v4.8.0 brought some serious changes, there are a lot of them in bindings too. I also took some cool ideas from @jonathanpeppers' glidex.

I'd like to stress the following changes:

1. Two Glide dependencies (GIF Decoder and Disk LRU Cache) have been added as separate packages since they now come in separate .aar and .jar respectively.
I added them as dependencies in NuGet spec of Xamarin.Android.Glide. However, to make it work, we need to manually add them directly to the main project where Glide is used. Otherwise, there's a runtime error as [docs say](https://bumptech.github.io/glide/doc/download-setup.html#download). I'm not sure whether I need to reference them from Xamarin.Android.Glide project too. Doing so doesn't fix the aforementioned error nor it's needed at compile time.

2. Unified project files and NuGet specs. Updated project files.

3. The original Glide package has a dependencies on Android Support Fragment and Animated Vector Drawable. However, in our NuGet spec we have dependencies on Android Support v4 and Annotations. I'm not sure what to do here.

Once done with this PR, I'd like to add more integration libraries for Glide like [OkHttp3](https://bumptech.github.io/glide/int/okhttp3.html) and [RecyclerView](https://bumptech.github.io/glide/int/recyclerview.html).


Android/Glide/source/Xamarin.Android.Glide/Additions/Additions.cs
```diff
@@ -1,417 +1,412 @@
...
 		}
-
 	}
 }
 
+namespace Bumptech.Glide.Load.Engine.Executor
+{
+    public partial class GlideExecutor
+    {
+        IList IExecutorService.InvokeAll(ICollection tasks) => InvokeAll(tasks.Cast<ICallable>().ToList()).ToList();
```
rubo, line 185: I'm not sure this is the best way for `ICollection` to `ICollection<T>` and `IList<T>` to `IList` conversion.

xyfoo:
@rubo  I've implemented the RecyclerView integration library v4.0 a few days ago (https://github.com/xyfoo/Bumptech.Glide.Integration.RecyclerView). Not sure if that would be useful to you.

rubo:
@xyfoo Thanks for sharing. I see you had a problem with bindings and went with pure C# instead. Nevertheless, it should be a binding project. I'll check once I'm done with this.

Redth:
Thanks, I merged this into another branch and just going to clean up a few minor things, the contribution is much appreciated!

PR #622, opened 2019-08-02
Set up CI with Azure Pipelines

mattleibow:


azure-pipelines.yml
```diff
@@ -0,0 +1,120 @@
+
+trigger:
+  - master
+
+variables:
+  MONO_VERSION: 5_18_1
+  XCODE_VERSION: 10.2.1
+  CAKE_NUGET_USEINPROCESSCLIENT: false
```
mattleibow, line 8: There seems to be some issues all of a sudden... Using this as an interim solution.

common.cake
```diff
@@ -7,7 +7,13 @@
...
 #addin nuget:?package=Newtonsoft.Json&version=9.0.1
 #addin nuget:?package=Cake.Json&version=3.0.1
-#addin nuget:?package=Mono.ApiTools.NuGetDiff&version=1.0.1&loaddependencies=true
+#addin nuget:?package=Mono.ApiTools.NuGetDiff&version=1.0.2-preview.4&loaddependencies=true
+
+// if using the out-of-process nuget client:
+#addin nuget:?package=NuGet.Packaging&version=4.6.0
+#addin nuget:?package=NuGet.Versioning&version=4.6.0
+#addin nuget:?package=NuGet.Frameworks&version=4.6.0
+#addin nuget:?package=NuGet.Packaging.Core&version=4.6.0
```
mattleibow, line 16: As a result of the interim solution of using the nuget.exe, we need this as well.

Redth (approved):

PR #629, opened 2019-08-07
Updated ARCore to 1.10.0

craigktreasure:
This change updates ARCore bindings to version 1.10.0 (based on #583). I have run the HelloAR sample app and verified that it works correctly.

I realize that ARCore 1.11.0 is out and will follow-up with that change as well, but I think it's valuable to have packages for 1.10.0 as well.

Since the SceneForm projects depend on the ARCore NuGet package, i'll have to follow up with updates to the SceneForm bindings separately once the ARCore 1.10.0 NuGet package is available.

### DIFF
--- a/Android/ARCore/build.cake
+++ b/Android/ARCore/build.cake
@@ -2,9 +2,9 @@
 
 var TARGET = Argument ("t", Argument ("target", "Default"));
 
-var NUGET_VERSION = "1.9.0";
+var NUGET_VERSION = "1.10.0";
 
-var AAR_VERSION = "1.9.0";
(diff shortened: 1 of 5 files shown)


craigktreasure:
@Redth and @newky2k Let me know if there's anything else you'd like me to validate and if there's anything else required on my part to get the updated package published to nuget.org.

Redth:
Had some CI issues where the API diffs between the update and existing nuget package weren't generating, so running again and will have a look tomorrow :)

Redth (approved):

Redth:
Excellent, API Diff looks good, thanks for this!

PR #636, opened 2019-08-09
Add NuGet version Suffix for prerelease packages

mattleibow:
Depends on https://github.com/xamarin/XamarinComponents/pull/628

### DIFF
--- a/Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
+++ b/Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
@@ -206,7 +206,8 @@ namespace AndroidBinderator
 				{
 					Name = mavenArtifact.ArtifactId,
 					NuGetPackageId = mavenArtifact.NugetPackageId,
-					NuGetVersion = mavenArtifact.NugetVersion,
+					NuGetVersionBase = mavenArtifact.NugetVersion,
+					NuGetVersionSuffix = config.NugetVersionSuffix,
 					MavenGroupId = mavenArtifact.GroupId,
 					AssemblyName = mavenArtifact.AssemblyName,
 					Config = config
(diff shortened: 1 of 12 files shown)


Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Model/BindingProjectModel.cs
```diff
@@ -14,7 +14,13 @@ public class BindingProjectModel
...
 
 		public string NuGetPackageId { get; set; }
-		public string NuGetVersion { get; set; }
+		public string NuGetVersionBase { get; set; }
+		public string NuGetVersionSuffix { get; set; }
+
+		public string NuGetVersion =>
+			string.IsNullOrWhiteSpace(NuGetVersionSuffix)
+				? NuGetVersionBase
+				: $"{NuGetVersionBase}-{NuGetVersionSuffix}";
```
mattleibow: This is technically a breaking change since the version is now readonly, and will change based on the suffix.

mattleibow: Also, should we be adding the `-`, or leave it up to the user?

Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
```diff
@@ -91,8 +94,6 @@ static async Task ProcessConfig(MavenRepository maven, BindingConfig config)
 					.UseMemoryCachingProvider()
 					.Build();
 
-				var slnFileInfo = new FileInfo(config.SolutionFile);
```
mattleibow, line 94: This didn't seem to be used...

Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Model/BindingProjectModel.cs
Redth: I think we leave `-` up to the user.  We need to not apply the suffix for dependencyOnly == true.

Redth (approved):

PR #641, opened 2019-08-13
Update Android Binderator

mattleibow:
The first change is just a basic one to add support for using Maven Central as a source.

The next (main) change is the support of "metadata" in the config file. Each level supports metadata:
 - Config file (global)
 - Templates
 - Artifacts

As you go down a level, you get all the parent values. So, the artifact metadata will have all three config, template and artifact metadata. The template metadata will only have the config and template metadata.

This is used in the form `config.json`:

```json
{
  ...
  "metadata": {
    "key": "value"
  },
  "templates": [
    {
    ...
    "metadata": {
      "key": "value"
    }
  }],
  "artifacts": [
    {
      ...
      "metadata": {
        "key": "value"
      }
    }
  ]
}
```

This is then consumed in the `.cshtml`:

```cshtml
@Model.Metadata["key"]
```


Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator.Tool/Xamarin.AndroidBinderator.Tool.csproj
```diff
@@ -22,12 +22,9 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
   </PropertyGroup>
-  
+
   <ItemGroup>
-    <PackageReference Include="MavenNet" Version="2.1.2" />
     <PackageReference Include="Mono.Options" Version="5.3.0.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
-    <PackageReference Include="RazorLight" Version="2.0.0-beta1" />
```
mattleibow, line 30: The the project reference pull these in to avoid warnings of downgrades.

Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Config/BindingConfig.cs
```diff
@@ -55,6 +58,7 @@ public enum MavenRepoType
 	{
 		Url,
 		Directory,
-		Google
+		Google,
+		MavenCentral
```
mattleibow, line 62: This is the new location.

Redth (approved):

PR #652, opened 2019-08-27
[Guava] Use .targets files for the jars

mattleibow:
Because the .dlls are not used at all, MSBuild thinks that it can discard it - taking the embedded .jars with it. 

Here we add a .targets file to always include the .jars, and this means that MSBuild can discard the .dlls as well as not have to extract the .jars from the .dlls - which is extra processing.

This is a fairly big change, not because of the .dll or the .jar, as those are still the same APIs, but rather since we are now using the .targets file instead of embedding the .jar in the .dll. As a result of this change, project that reference the package no longer get the .jar via transitive properties.
What this means is that if I have a project, MyAndroidLibrary which uses Guava, and then MyAndroidApp that references MyAndroidLibrary, I actually won't get the .jar because the default settings in NuGet do not include .targets files, which means that MSBuild never knows to include the .jar.

In most cases, this is not such a big deal as we can always install the Guava package into the main app project as well, then everything works as expected.

What this does mean for library/package authors is that they have to specify `PrivateAssets="None"` on the `<PackageReference>`:

```xml
<PackageReference Include="Xamarin.Google.Guava" Version="27.1.0.2" PrivateAssets="None" />
```

This will cause the MSBuild `Pack` target to generate the .nuspec file with the correct dependency definition:

```xml
<dependency id="Xamarin.Google.Guava.ListenableFuture" version="1.0.0.1" include="All" />
```

You can't also use `ExcludeAssets` in the `PackageReference` as this affects the including in the _current_ project, not the generated dependencies.

Interestingly, when using the .nuspec directly with `nuget pack` (without specifying the `include` and `exclude` attributes), the default is to include everything - including the .targets files... 

One thing to note, is that if we specify `PrivateAssets="None"` in the reference, the actual .targets file will run. So, if we have a binding project, and the targets references a .jar with a build action of `<EmbeddedJar>`, then that jar from the referenced package will become embedded into the  binding project. The output package file will then have the jar embedded, and referenced via a dependency - resulting in duplicate classes. 

For this case, this is not a problem since we are carefully controlling what we are doing. We are using the `<AndroidJavaLibrary>` build action, which does not affect binding projects nor class library projects. This build action is just for app projects that are at the top of the hierarchy - and are safe to use.

This will all become nice and easy once this is implemented: https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior


Android/Guava/build.cake
```diff
@@ -122,6 +122,7 @@ Task("nuget")
 		c.Properties.Add("PackageOutputPath", new [] { MakeAbsolute(new FilePath("./output")).FullPath });
 		c.Properties.Add("PackageRequireLicenseAcceptance", new [] { "true" });
 		c.Properties.Add("DesignTimeBuild", new [] { "false" });
+		c.Properties.Add("NoBuild", new [] { "true" });
```
mattleibow, line 125: No need to rebuild the projects if the libs task builds it.

Android/Guava/source/Guava.FailureAccess/Guava.FailureAccess.csproj
```diff
@@ -20,38 +20,30 @@
...
-    <None Include="..\..\externals\guava-failureaccess.jar">
-      <Link>guava-failureaccess.jar</Link>
-    </None>
+    <None Include="Guava.FailureAccess.targets" Pack="True" PackagePath="build\$(TargetFramework)\$(PackageId).targets" />
+    <None Include="..\..\externals\guava-failureaccess.jar" Pack="True" PackagePath="build\$(TargetFramework)\guava-failureaccess.jar" />
     <None Include="..\..\External-Dependency-Info.txt" Pack="True" PackagePath="THIRD-PARTY-NOTICES.txt" />
   </ItemGroup>
 
-  <ItemGroup>
-    <EmbeddedJar Include="..\..\externals\guava-failureaccess.jar" />
```
mattleibow, line 49: We can't embed the jar as Release builds make sure to remove all assemblies that are not referenced, and this assembly will never be referenced - it can't be as there are no types to reference.
Also, no need to extract the jar, that is just extra work.

Android/Guava/source/Guava.FailureAccess/Guava.FailureAccess.csproj
```diff
@@ -20,38 +20,30 @@
...
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=865028</PackageProjectUrl>
     <PackageLicenseUrl>https://go.microsoft.com/fwlink/?linkid=865030</PackageLicenseUrl>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageVersion>1.0.1</PackageVersion>
+    <PackageVersion>1.0.1.1</PackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
-    <TransformFile Include="Transforms\Metadata.xml" />
```
mattleibow, line 38: We don't need any of the transforms files since we are not actually giving any .jars to the binding generator.

Android/Guava/source/Guava.FailureAccess/Guava.FailureAccess.csproj
```diff
@@ -20,38 +20,30 @@
...
     <Summary>Xamarin.Android bindings for Google Guava FailureAccess</Summary>
     <Description>
-    NOTE: This package is meant to be used as a reference only for other binding projects that depend on Guava.
-    This package does not expose any useful API's in the .dll itself.
-    
-    Guava is a set of core libraries that includes new collection types (such as multimap and multiset), immutable collections, a graph library, functional types, an in-memory cache, and APIs/utilities for concurrency, I/O, hashing, primitives, reflection, string processing, and much more!
+NOTE: This package is meant to be used as a reference only for other binding projects that depend on Guava.
+This package does not expose any useful API's in the .dll itself.
+
+Guava is a set of core libraries that includes new collection types (such as multimap and multiset), immutable collections, a graph library, functional types, an in-memory cache, and APIs/utilities for concurrency, I/O, hashing, primitives, reflection, string processing, and much more!
```
mattleibow, line 26: Even though indents make the .csproj look nice, it messes up the nuget.org page: 
image

Android/Guava/source/Guava.FailureAccess/Guava.FailureAccess.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>MonoAndroid50</TargetFrameworks>
+    <TargetFramework>MonoAndroid50</TargetFramework>
```
mattleibow, line 4: We don't need the plural if we only have one.

newky2k (approved):

PR #670, opened 2019-09-17
upgrade InAppSettingsKit to 2.12

newky2k:
Pulled into Xamarin branch


newky2k:
@Wenfengcheng Looks like there are some issues when building the sample which is why its failing on CI.  You should try and open the samples/InAppSettingsKitSample.sln file and build it to see what they are.

thanks for you contribution so far

PR #677, opened 2019-09-30
[Xamarin.Kotlin.BindingSupport] Make sure to use properly formed URLs

mattleibow:


Util/Xamarin.Kotlin.BindingSupport/native/src/main/kotlin/xamarin/binding/kotlin/bindingsupport/Processor.kt
```diff
@@ -49,7 +49,7 @@ class Processor(xmlFile: File, jarFiles: List<File>, outputFile: File?) {
     private val xtransformsdoc: Document
 
     init {
-        val urls = jarFiles.map { file -> URL("file://${file.canonicalPath}") }
+        val urls = jarFiles.map { file -> file.toURI().toURL() }
```
mattleibow, line 52: This was not cross-platform enough. Seems no exceptions are thrown if the URL is in an invalid format.

Redth (approved):

PR #679, opened 2019-09-30
Update Android Binderator to include .NET Core 3.0

mattleibow:
Update the Android Binderator.

Once merged, we merge the PR to use this new tool - and update the .ci/build.yml to use all the new tooling versions natively: https://github.com/xamarin/XamarinComponents/pull/680


Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator.Tool/Xamarin.AndroidBinderator.Tool.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <PackAsTool>true</PackAsTool>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
```
mattleibow, line 6: Build to support .NET Core 2.1+ and .NET Core 3.0+

azure-pipelines.yml
```diff
@@ -18,6 +18,11 @@ jobs:
     parameters:
       timeoutInMinutes: 360
       buildType: 'manifest'
+      mono: '6_4_0'
+      xcode: '11'
+      dotnet: '3.0.x'
+      cake: '0.35.0'
+      apiTools: '1.1.0-preview.1'
```
mattleibow, line 25: Temporarily override the versions so we can build. This will be removed once the tooling can run on .NET Core 3.0.

We need to first release the updated binderator so that it runs nicely on .NET Core 3.0. Once that is done, then we can use it. If we update the `.ci/build.yml` here, then Android Support will fail to build because the current version of binderator will say that the tool does not support .NET Core 3.0.

azure-pipelines.yml
```diff
@@ -18,6 +18,11 @@ jobs:
     parameters:
       timeoutInMinutes: 360
       buildType: 'manifest'
+      mono: '6_4_0'
+      xcode: '11'
+      dotnet: '3.0.x'
+      cake: '0.35.0'
```
Redth, line 24: Is cake ok for 3.0?

mattleibow: Yep, with the brand spanking new 0.35!

Redth (approved):

PR #680, opened 2019-09-30
Update build tooling versions

mattleibow:
Use new versions.

This can only be merged after the PR to build the new tools is released: https://github.com/xamarin/XamarinComponents/pull/679

- [x] We also need to update the main yaml to not override the yaml template.


azure-pipelines.yml
```diff
@@ -12,20 +12,16 @@ resources:
       type: github
       name: xamarin/XamarinComponents
       endpoint: xamarin
+      ref: refs/heads/dev/update-tooling
```
mattleibow: REMOVE LINE BEFORE MERGING

PR #682, opened 2019-10-02
Bump renderscriptv8 to 28.0.0.3

Redth:

PR #684, opened 2019-10-02
Constraint Layout 2.0.0-beta2 Update

Redth:


Android/ConstraintLayout/build.cake
```diff
@@ -1,35 +1,65 @@
...
 var CONSTRAINT_LAYOUT_SOLVER_URL = string.Format ("https://dl.google.com/dl/android/maven2/com/android/support/constraint/constraint-layout-solver/{0}/constraint-layout-solver-{0}.jar", AAR_VERSION);
 
+Task("libs")
+	.IsDependentOn("externals")
+	.Does (() =>
+{
+	MSBuild("./source/ConstraintLayout.sln", c => {
+		c.Configuration = "Release";
+		c.Targets.Clear();
+		c.Targets.Add("Restore");
```
mattleibow: Maybe rather use the `c.Restore = true;` and then don't worry about specifying the targets (ie: just use the default build)

Android/ConstraintLayout/build.cake
```diff
@@ -1,35 +1,65 @@
...
+});
+
+Task("samples")
+	.IsDependentOn("nuget")
+	.Does (() =>
+{
+	MSBuild("./samples/ConstraintLayoutSample.sln", c => {
+		c.Configuration = "Release";
+		c.Targets.Clear();
+		c.Targets.Add("Restore");
```
mattleibow: Same as above, just use `c.Restore = true;`

Android/ConstraintLayout/build.cake
```diff
@@ -1,35 +1,65 @@
...
-		new DefaultSolutionBuilder { SolutionPath = "./samples/ConstraintLayoutSample.sln", BuildsOn = BuildPlatforms.Windows | BuildPlatforms.Mac },
-	},
+Task("nuget")
+	.IsDependentOn("libs")
+	.Does (() =>
+{
+	MSBuild ("./source/ConstraintLayout.sln", c => {
+		c.Configuration = "Release";
+		c.Targets.Clear();
+		c.Targets.Add("Pack");
```
mattleibow, line 50: One thing that you can do to speed up the pack and not trigger a build (because you just built it already) is to add the `NoBuild` property: `c.Properties.Add("NoBuild", new [] { "true" });`

manifest.yaml
```diff
@@ -17,7 +17,8 @@
 - Name: ConstraintLayout
   BuildScript: ./Android/ConstraintLayout/build.cake
   TriggerPaths: [ Android/ConstraintLayout ]
-  MacBuildTargets: [ nuget, samples ]
+  MacBuildTargets: [ samples ]
+  WindowsBuildTargets: [ samples ]
```
mattleibow: Maybe use `ci` here to slowly start the move towards a manifest-less world.

Android/ConstraintLayout/build.cake
```diff
@@ -1,35 +1,65 @@
-#load "../../common.cake"
+var TARGET = Argument ("target", Argument ("t", "ci"));
 
-var TARGET = Argument ("target", Argument ("t", "Default"));
-
-var NUGET_VERSION = "2.0.0-alpha2";
-var AAR_VERSION = "2.0.0-alpha2";
+var NUGET_VERSION = "2.0.0-beta2";
+var AAR_VERSION = "2.0.0-beta2";
```
mattleibow, line 4: I see the version numbers haven't changed, did we just not release it?

Redth: -alpha2 vs -beta2...

Android/ConstraintLayout/build.cake
Redth: It really doesn't add much time to it, I think we should KISS and leave as is, but good to know!

Android/ConstraintLayout/samples/NuGet.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <packageSources>
+        <add key="BuildOutput" value="../output" />
```
mattleibow, line 4: Might want to do the same here: https://github.com/xamarin/XamarinComponents/pull/682/files#r330617947

mattleibow (approved):

mattleibow (approved):
LGTM

PR #687, opened 2019-10-03
Master based grpc 1.21

moljac:


Util/Xamarin.Kotlin.BindingSupport/native/.idea/workspace.xml
```diff
@@ -1,110 +0,0 @@
...
-          <list>
-            <option value="jar" />
-          </list>
-        </option>
-        <option name="vmOptions" />
-      </ExternalSystemSettings>
-      <GradleScriptDebugEnabled>true</GradleScriptDebugEnabled>
-      <method v="2" />
-    </configuration>
-    <configuration name="xamarin.binding.kotlin.bindingsupport.ProgramKt" type="JetRunConfigurationType" factoryName="Kotlin" nameIsGenerated="true">
```
mattleibow: Could we keep these .idea files so that the dev does not have to configure the run targets for the app or the build target for the jar?
It somehow feels like it should be removed, but if you do, then intellij suddenly does not understand how to run the kotlin app.

XPlat/Google.Grpc/build.cake
```diff
@@ -80,7 +41,7 @@ Dictionary<string, string> JAR_URLS_ARTIFACT_FILES= new Dictionary<string, strin
 	},
 };
 string ARTIFACT_FILE = "";
-string NUGET_VERSION=$"{ARTIFACT_VERSION}.2";
+string NUGET_VERSION=$"{ARTIFACT_VERSION}";
 
 
 BuildSpec buildSpec = new BuildSpec ()
```
mattleibow, line 47: We should probably move away from this BuildSpec and to standard cake: https://github.com/xamarin/XamarinComponents/pull/684/files#diff-c578a2377ba3bb7693c87d13619d81c4

XPlat/Google.Grpc/source/Xamarin.Grpc.Android.Bindings.XamarinAndroid/Xamarin.Grpc.Android.Bindings.XamarinAndroid.csproj
```diff
@@ -0,0 +1,100 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <!--
+  https://montemagno.com/converting-xamarin-libraries-to-sdk-style-multi-targeted-projects/
+    
+  -->
+    <PropertyGroup>
+        <AssemblyName>Xamarin.Grpc.Android</AssemblyName>
```
mattleibow, line 7: Because this project already has the Android in the name, we might as well leave off the `.Bindings.XamarinAndroid` since this will ALWAYS be a binding and Android is in the name. Same with the rest.

This project can just be named `Xamarin.Grpc.Android`. It will also be quite a few less characters.

moljac: If you check you will see that pattern was:

for namespaces:
```
Xamarin.Grpc.Core
Xamarin.Grpc.Context
Xamarin.Grpc.Stub
Xamarin.Grpc.OkHttp
Xamarin.Grpc.Protobuf.Lite
```
and for projectnames - add sufix `Bindings.XamarinAndroid`

Why `Bindings`? I because I have templates where I have non-bindings projects for pure Xamarin.Android platform specific code. (I know that can go to `./Additions/`, but it was template decision choice). And if I need it I simply delete project generated by template.

Why `XamarinAndroid`? To avoid issues if namespace is `Android`.

And now `Android`. Google added new Grps artifact `grpc-android`.

https://mvnrepository.com/artifact/io.grpc/grpc-android

yes. I'm aware this might force users into using FQ typenames with prefix `global::`, but it was google that added artifact with the name that is not ideal for us.

XPlat/Google.Grpc/build.cake
moljac: I moved away from BuildSpec to standard cake on all of my PRIVATE projects. initially i wanted to have the same Cake "framework" for Xamarin/microsoft and my stuff, but the fight with keeping Cake work with .NET core, mono versions and various libraries it was too much wasted time. So I switched to standard cake.

I had few bindings here and in other repos with vanilla Cake, but they were rewritten to BuildSpec, so I understood that `BuildSpec` is way to go.

Util/Xamarin.Kotlin.BindingSupport/native/.idea/workspace.xml
moljac: I do not know how those files got into master. They popped up suddenly 2-3 weeks ago and come from `Xamarin.Kotlin.BindingsSupport`. I was locally dragging it for weeks and had to take care of it.

All my PRIVATE project have `**/.idea` in `.gitignore`. Rider folder. Shouldn't we add those to `.gitignore`?

moljac: I assumed those files slipped in by accident. My 1st decision was to keep em and take care not to commit or do anything. 
After few weeks I changed my mind and decided to remove them, though they were not mine. Maybe that was the mistake. Sorry.

XPlat/Google.Grpc/build.cake
moljac: I'm just trying to bump the bindings, so I can continue working on AndroidX updates.

Let's talk about vanilla Cake in the meeting and decide what is the best and after decision I will rewrite scripts.

IMO moving away from BuildSpecs is good idea. But I d like to finish some other priorities 1st.

XPlat/Google.Grpc/source/Xamarin.Grpc.Android.Bindings.XamarinAndroid/Xamarin.Grpc.Android.Bindings.XamarinAndroid.csproj
moljac: Please, leave this less important style issue as is. Let us agree on naming and I'll obey. Check Jon's reviews week or two ago - this was not the issue.

New artifact with unlucky name came in, but changing namespace names would be breaking change.

Changing projectnames is low priority for me. Unless Hugo.Boss has different opinion.

Util/Xamarin.Kotlin.BindingSupport/native/.idea/workspace.xml
Redth: Yeah sounds like we should bring them back, can you revert the deletes of those please?

Redth (approved):

PR #685, opened 2019-10-02
Improve Kotlin Binding Support

mattleibow:

PR #689, opened 2019-10-04
Improve Kotlin Binding Support

mattleibow:
This is a remake of https://github.com/xamarin/XamarinComponents/pull/685

For some reason the PR passed, but the merge to master fails consistently. Something about duplicate kotlin methods... Only on Windows, and sometimes mac locally... Sounds legit.

### DIFF
--- a/Util/Xamarin.Kotlin.BindingSupport/build.cake
+++ b/Util/Xamarin.Kotlin.BindingSupport/build.cake
@@ -1,7 +1,7 @@
 
 var TARGET = Argument("t", Argument("target", "Default"));
 
-var PACKAGE_VERSION = "0.2.0-preview";
+var PACKAGE_VERSION = "0.3.0-preview";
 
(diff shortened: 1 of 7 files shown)


mattleibow:
After all that, still no idea why it fails sometimes. A rerun makes it work. A rebuild makes it work. I can also do a git clean and it will work.

The part that is failing is basically:

https://github.com/xamarin/XamarinComponents/blob/46153434590ea5324a7bc795bd3cf16eaf6df924/Util/Xamarin.Kotlin.BindingSupport/native/src/main/kotlin/xamarin/binding/kotlin/bindingsupport/Processor.kt#L354

Sometimes Kotlin can't locate the member - when I have it already...

This is a bit worrying in that I am not sure what else is going wrong, but at least we can ignore the errors and move on.

Redth (approved):

PR #696, opened 2019-10-07
Make Xamarin.Build.Download FIPS Compliant

Redth:


Util/Xamarin.Build.Download/build.cake
```diff
@@ -1,28 +1,54 @@
...
+	MSBuild ("./source/Xamarin.Build.Download.sln", c => {
+		c.Configuration = "Release";
+		c.Restore = true;
+		c.Targets.Clear();
+		c.Targets.Add("Pack");
+		c.Properties.Add("PackageOutputPath", new [] { MakeAbsolute(new FilePath("./output")).FullPath });
+		if (!string.IsNullOrEmpty(SOURCE_BRANCH))
+			c.Properties.Add("RepositoryBranch", new [] { SOURCE_BRANCH });
+		if (!string.IsNullOrEmpty(SOURCE_COMMIT))
+			c.Properties.Add("RepositoryCommit", new [] { SOURCE_COMMIT });
```
mattleibow: Since we may actually want to force this on all builds, we could do this in the template:

https://github.com/xamarin/XamarinComponents/blob/master/.ci/build.yml#L187-L188

Util/Xamarin.Build.Download/source/Xamarin.Build.Download.Tests/HashTests.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+using Xamarin.Build.Download;
+using Xunit;
+
+namespace NativeLibraryDownloaderTests
+{
+    public class HashTests
+    {
+        [Theory]
+        [InlineData("C:\\path\\to\\file.aar")]
```
mattleibow: This will fail on mac with invalid paths. I think?

Util/Xamarin.Build.Download/source/Xamarin.Build.Download.Tests/Helpers/TestsBase.cs
```diff
@@ -28,8 +27,10 @@ public void SetUp ()
 			Environment.CurrentDirectory = tempDir;
 		}
 
-		[OneTimeTearDown]
-		public void TearDown ()
+        public bool IsWindows
```
mattleibow: I know this is just text, but I still hope the files will have either tabs or spaces.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildGetArchivesToDownload.cs
```diff
@@ -57,8 +57,8 @@ public override bool Execute ()
 					taskItem.SetMetadata ("RangeStart", partialZipDownload.RangeStart.ToString ());
 					taskItem.SetMetadata ("RangeEnd", partialZipDownload.RangeEnd.ToString ());
 
-					if (!string.IsNullOrEmpty (partialZipDownload.Md5))
-						taskItem.SetMetadata ("Md5", partialZipDownload.Md5);
+					if (!string.IsNullOrEmpty (partialZipDownload.Sha256))
+						taskItem.SetMetadata ("Sha256", partialZipDownload.Sha256);
```
mattleibow, line 61: I see all these nice changes, but are we backwards compatible?

Util/Xamarin.Build.Download/source/Xamarin.Build.Download.Tests/HashTests.cs
Redth: The intent is just to see that i get no funny characters (just numbers, and digits) in the CRC code returned.  I'll add a test case for a mac path too for kicks.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinBuildGetArchivesToDownload.cs
Redth: It should just skip over verifying the old md5 hashes since it no longer checks for them and no longer sets the task item, it will just never be checked to be validated.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download.Tests/Test.cs
```diff
@@ -471,7 +487,6 @@ public void TestDisallowUnsafeGetItemsToDownload ()
 					{ "ToFile", "cardview.v7.aar" },
 					{ "RangeStart", "196438127" },
 					{ "RangeEnd", "196460160" },
-					{ "Md5", "b44eb88f7cc621ae616744c6646f5b64" }
```
mattleibow, line 474: We are removing a check to make sure that the file is both correct/complete as well as the right file. We need to have a new way to do this. Since maven only has SHA1 and MD5, we are probably going to have to do the hashing ourselves on CI or as part of the binding process.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.csproj
```diff
@@ -1,94 +1,41 @@
...
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="packages.config" />
-    <None Include="Xamarin.Build.Download.props">
+    <None Include="Xamarin.Build.Download.props" Pack="True" PackagePath="build\Xamarin.Build.Download.props">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="$(TargetPath)" Pack="True" PackagePath="build\Xamarin.Build.Download.dll" />
+    <None Include="$(PkgMono_Cecil)\lib\netstandard2.0\Mono.Cecil.dll" Visible="False" Pack="True" PackagePath="build\Mono.Cecil.dll" />
+    <None Include="$(PkgNewtonsoft_Json)\lib\netstandard2.0\Newtonsoft.Json.dll" Visible="False" Pack="True" PackagePath="build\Newtonsoft.Json.dll" />
```
mattleibow, line 32: I like this.

manifest.yaml
```diff
@@ -592,7 +592,8 @@
 - Name: Xamarin.Build.Download
   BuildScript: ./Util/Xamarin.Build.Download/build.cake
   TriggerPaths: [ Util/Xamarin.Build.Download ]
-  MacBuildTargets: [ nuget ]
+  MacBuildTargets: [ ci ]
+  WindowsBuildTargets: [ ci ]
```
mattleibow, line 616: I like this pattern of adding both the Windows and Mac agents in. Should start showing any cross-platform issues before it even gets to the gate.

mattleibow (approved):
Looking good. 2 tests seem to be failing.

mattleibow (approved):
LGTM

PR #694, opened 2019-10-06
First Implementation of Google Dagger for Android

Iguins:
Google Dagger is required for Firebase InAppMessaging in Xamarin GooglePlayServiceComponent Rep.

PR #697, opened 2019-10-07
Final Google Dagger fixes

Redth:
There's some house keeping we need to do for this (originally PR #694 - thanks @GuillaumeSE !):

 - [x] Create FWLinks
-  [x] Add cake ci target
 - [x] Add to manifest


Android/GoogleDagger/source/Dagger/Dagger.csproj
```diff
@@ -0,0 +1,49 @@
...
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageVersion>2.13</PackageVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <TransformFile Include="Transforms\*.xml" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="..\..\externals\dagger.jar">
```
mattleibow: Why is this included here, and then again below?

Android/GoogleDagger/source/Dagger/Dagger.csproj
```diff
@@ -0,0 +1,49 @@
...
+    <Owners>Microsoft</Owners>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=2106537</PackageProjectUrl>
+    <PackageLicenseUrl>https://go.microsoft.com/fwlink/?linkid=2106538</PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageVersion>2.13</PackageVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <TransformFile Include="Transforms\*.xml" />
```
mattleibow, line 32: This actually results in duplicate references to the xml files. I would suggest adding

```
<None Remove="Transforms\*.xml" />
```

above the line so that we first remove the xml files, and then add them back as the real thing.

manifest.yaml
```diff
@@ -224,6 +229,10 @@
   BuildScript: ./Android/GoogleBillingClient/build.cake
   TriggerPaths: [ Android/GoogleBillingClient ]
   MacBuildTargets: [ libs, nuget]
+- Name: Dagger
```
mattleibow: I don't think we need this as well, just the GoogleDagger above should be good enough.

mattleibow (changes requested):
Almost there! Just a few tweaks.

mattleibow (approved):
Good to GO!

Issue #693, opened 2019-10-04
Update BillingClient to latest Jar version


EmilAlipiev:
According to the release notes, google solved 2 important issues as below. can we update this library using 2.0.3? i think that it is still on 2.0.0

 https://developer.android.com/google/play/billing/billing_library_releases_notes

Bug fixes
Fixed a bug where querySkuDetailsAsync() would occasionally fail with code DEVELOPER_ERROR instead of returning a successful result.

Bug fixes
Fixed a bug where debug messages were being returned as null in some cases.
Fixed a potential memory leak issue.

PR #705, opened 2019-10-09
Upgrade Google BillingClient to v2.0.3

tuyen-vuduc:
#693 Upgrade Google BillingClient t0 v2.0.3


Android/GoogleBillingClient/source/GoogleBillingClient/GoogleBillingClient.csproj
```diff
@@ -23,11 +24,15 @@
...
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=2100110</PackageProjectUrl>
-    <PackageLicenseUrl>https://go.microsoft.com/fwlink/?linkid=2099698</PackageLicenseUrl>
+    <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageVersion>2.0.0</PackageVersion>
+    <PackageVersion>2.0.3</PackageVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Include="../../LICENSE.md" Pack="true" PackagePath="LICENSE.md"/>
```
Redth: The actual source is `../../License.md` which may matter on macOS, could you please update it here?

Redth: The actual source is ../../License.md which may matter on macOS, could you please update it here?
```suggestion
    <None Include="../../License.md" Pack="true" PackagePath="LICENSE.md"/>
```

tuyen-vuduc: Thank you. I made the change and also change to backslash following other references.

tuyen-vuduc:
@Redth Is it ok to merge?

Redth:
@tuyen-vuduc trying to see if there's any sane way to not break the few api methods here:

```
# API diff: Xamarin.Android.Google.BillingClient.dll

## Xamarin.Android.Google.BillingClient.dll

### Namespace Android.BillingClient.Util

#### Type Changed: Android.BillingClient.Util.BillingHelper

Removed methods:

public static Android.OS.Bundle ConstructExtraParamsForAcknowledgePurchase (Android.BillingClient.Api.AcknowledgePurchaseParams acknowledgePurchaseParams);
public static Android.OS.Bundle ConstructExtraParamsForConsume (Android.BillingClient.Api.ConsumeParams consumeParams, bool isIabV9Supported);
public static Android.OS.Bundle ConstructExtraParamsForGetSkuDetails (bool isIabV9Supported, bool enablePendingPurchases);
```

You'll notice Google added `string libraryVersion` parameters to these methods, and I'm not sure if passing `null` is acceptable or not.  If it is, we could add a partial class with overloads for these methods which call the new ones just passing `null` as the `libraryVersion`.  The documentation doesn't even yet reflect these API changes unfortunately.

tuyen-vuduc:
@Redth 

The above methods I think they are intended to use internally within the library. If they were changed for some particular reason, I think we should keep them as they are.

If someone already uses one or some of them, they should change their code after the upgrade to ensure it works as expect. It's a must in the native world, obviously.

How do you think?

Redth (approved):

PR #708, opened 2019-10-11
Create a script to build from directories

mattleibow:
This is very similar to the manifest, but instead of a yaml file, we create a simpler structure based on the locations of build.cake files


.ci/build-directories.cake
```diff
@@ -0,0 +1,366 @@
...
+
+var BUILD_NAMES = Argument("names", Argument("name", Argument("n", "")))
+	.Split(new [] { ",", ";" }, StringSplitOptions.RemoveEmptyEntries);
+var BUILD_TARGETS = Argument("targets", Argument("build-targets", Argument("build-targets", Argument("build", DEFAULT_BUILD_TARGET))))
+	.Split(new [] { ",", ";" }, StringSplitOptions.RemoveEmptyEntries);
+
+var FORCE_BUILD = Argument("force", Argument("forcebuild", Argument("force-build", false)));
+var POD_REPO_UPDATE = Argument("update", Argument("repo-update", Argument("pod-repo-update", false)));
+
+var BUILD_GROUPS = GetFiles($"{ROOT_DIR}/*/**/build.cake")
```
mattleibow, line 31: Instead of the yaml file, our script is based on these build.cake files. As a result the component "name" is now the path to the folder with the cake.

For example, the "Kotlin" component is now "Android/Kotlin". It always builds on both Windows and macOS, so the build script has to decide wether or not to run anything.

.ci/build-directories.cake
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// If a filter was specified, then use it
+if (BUILD_NAMES.Length > 0) {
+	groupsToBuild = groupsToBuild
+		.Where(bg => BUILD_NAMES.Any(n =>
+			// match exact "Android/Kotlin"
+			bg.FullPath.Equals(n, StringComparison.OrdinalIgnoreCase) ||
+			// match partial "Kotlin"
+			bg.FullPath.Replace("\\", "/").EndsWith("/" + n, StringComparison.OrdinalIgnoreCase)))
```
mattleibow, line 162: Instead of requiring the full directory path, such as "Android/Kotlin", you can also use the simple name of "Kotlin". This is to help with simplifying tags and other things. If there is any ambiguity, then we can fall back to the full name.

azure-pipelines.yml
```diff
@@ -12,12 +12,14 @@ resources:
...
       endpoint: xamarin
+      ref: refs/heads/dev/remove-manifest
 
 jobs:
   - template: .ci/build.yml@components
     parameters:
       timeoutInMinutes: 360
-      buildType: 'manifest'
+      buildType: 'directories'
+      targetsFilter: 'ci'
```
mattleibow: The new target that will be built will be `ci`, so most probably any older builds will fail. But, this is not really an issue as they will only be building if they were changed - and if they are changed, then they need to be migrated.

Mac/ITSwitch/build.cake
```diff
@@ -1,35 +1,36 @@
...
 
-	Components = new [] {
-		new Component { ManifestDirectory = "./component", BuildsOn = BuildPlatforms.Mac},
-	},
-};
+Task("samples")
+	.IsDependentOn("nuget")
+	.Does(() =>
+{
+	MSBuild("./samples/ITSwitchSample/ITSwitchSample.sln", new MSBuildSettings()
```
mattleibow, line 26: Crazy to live in a world where macOS samples build on Windows, but here we are! No need for any fancy don't-build-on-Windows logic.

azure-pipelines.yml
```diff
@@ -12,12 +12,14 @@ resources:
       type: github
       name: xamarin/XamarinComponents
       endpoint: xamarin
+      ref: refs/heads/dev/remove-manifest
```
mattleibow: **!!! THIS LINE MUST BE DELETE BEFORE MERGING !!!**

If this is not removed, then master will fail as this branch is not there. If this is removed now, then the PR will fail as DevOps will use the master branch to download the template, which does not have this new build pattern.

mattleibow:
Changed the PR to not actually change existing functionality but just add an alternative build type.

I will use it with some other repos, and then later switch this repo after some testing.

PR #716, opened 2019-10-15
[xbd] VS2019 build tools do not contain 7-zip, use local installation instead

Barokai:
I encountered that 7-zip is not installed when building with visual studio 2019 build tools, therefore i've added a fallback similar to https://github.com/xamarin/XamarinComponents/pull/353 which uses local version of 7-zip


Barokai:
Just saw that https://github.com/xamarin/XamarinComponents/pull/696/commits/4ec9c01cac70f4dcf2d70d43d7f1876783c9d032 would move the handling of the 7zip executable and should be adapted too if this will be merged :)

Redth:
We still do ship 7zip with the Xamarin workload in Visual Studio 2019, so I'm not sure this is necessary.

Redth:
I also added some fallback checks to look for the local machine install if we can't find it anywhere else, which should address this case.  Thanks!

PawKanarek:
@Redth  But 7zip is not shipped with Build Tools 2019, so i had to add it manulally to path ```(VsInstallRoot)\Common7\IDE\Extensions\Xamarin\VisualStudio\7-Zip\```  
that line form this closed PR would save me a day... 

And it would be nice ass-saver for Error: ```C:\Users\Jenkins\.nuget\packages\xamarin.build.download\0.10.0\buildTransitive\Xamarin.Build.Download.targets(52,3): error : Could not find 7zip.exe in Xamarin installation [e:\JW\myApp\src\MyApp.App.XF.iOS\MyApp.App.XF.iOS.csproj ```

Oh dear, that was messy day...

PawKanarek:
@Redth On my Jenkins machine i don't need to instal whole visual studio, i need only build tools and i think that im not the only one there.

kcrg:
@Redth @PawKanarek Same here... Whole day wasted...

PR #727, opened 2019-10-30
Updated ARCore to 1.12.0

craigktreasure:
- Updates ARCore bindings to version 1.12.0.
- Fixed ARCore NuGet warnings: NU5125, NU5128
  - Fixed [NU5125](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5125) by including the license file in the nupkg.
  - Fixed [NU5128](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5128) by adding a dependency group for MonoAndroid7.0 to the nuspec.
- "Fixed" the cake build for ARCore
  - I made the cake build work similar to the SceneForm project, which appears to be a more modern formula and happens to also work on Windows (the existing formula did not). This makes the instructions the same on both macOS and Windows.
- Updated the ARCore README.md
  - This change updates the README with updated instructions on how to build the ARCore component and samples.

I have run the HelloAR sample app and verified that it works correctly.

I realize that ARCore 1.13.0 is out and will follow-up with that change as well, but I think it's valuable to have packages for 1.11.0 as well.

Since the SceneForm projects depend on the ARCore NuGet package, i'll have to follow up with updates to the SceneForm bindings separately once the ARCore 1.12.0 NuGet package is available.

### DIFF
--- a/Android/ARCore/README.md
+++ b/Android/ARCore/README.md
@@ -6,17 +6,46 @@ With ARCore, shape brand new experiences that seamlessly blend the digital and p
 
 ## Building
 
-Use the `build.cake` script to build.  You can execute it via one of the bootstrappers (`build.sh` on mac or `build.ps1` on windows):
+### Prerequisites
 
-Mac:
-```
-sh ../../build.sh --target libs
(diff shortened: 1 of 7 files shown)


craigktreasure:
@Redth

Redth (approved):

PR #736, opened 2019-11-13
Add some required checks

mattleibow:
This PR adds some of the required security checks. 

This is not complete as some of the checks need to run as part of an MSBuild operation. As a result, the current format of a separate job is incorrect. That will need some more design - especially since some files for Components Governance are generated at compile time. This is mainly to get the basic CredScan and PoliCheck similar running, as well as to just hook up everyone to TSA.


mattleibow:
Squash merge this bad boy!

PR #768, opened 2019-12-16
Update Google Android Vending Library

Redth:


mattleibow (commented):
Looks good so far. Obviously CI does not agree :)

mattleibow:
I see the interface `IPolicy` got a new method:

```
string LicensingUrl { get; }
```

May need to do something about that with versions...

Redth:
> I see the interface `IPolicy` got a new method:
> 
> ```
> string LicensingUrl { get; }
> ```
> 
> May need to do something about that with versions...

Not sure there's a lot we can or should do here.  I think this is a pretty niche library that is likely mostly directly used so devs will have to just update their implementations of `IPolicy` if that's something they have.

PR #774, opened 2019-12-20
Update XBD ItemGroup name for AndroidAarLibrary

Redth:
Some older nuget packages use the old itemgroup name but may also embed the .aar in the .dll.  This is now causing duplicate .jar issues at compilation for some users.

This fix is a breaking change, but in practice the old item group is not really used, the only case we might find is if someone in the wild is using this functionality.  The workaround in that case is to revert to an older version of XBD until the library using the old item group migrates to the new one.


mattleibow (approved):
Seems legit.

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.targets
```diff
@@ -92,9 +92,18 @@
 		</ItemGroup>
 	</Target>
 
-	<Target Name="_XamarinBuildDownloadAarInclude"
+	<Target Name="_XamarinBuildDownloadAarRestore"
 		Condition="'@(XamarinBuildDownloadRestoreAssemblyAar)'!=''"
 		DependsOnTargets="_XamarinBuildDownload">
+		<Warning
+			Code="XBD100"
+			Text="ItemGroup `XamarinBuildDownloadRestoreAssemblyAar` is no longer supported.  If your build is failing, revert to an older Xamarin.Build.Download version, or migrate your packages to use the new `XamarinBuildDownloadAndroidAarLibrary` ItemGroup.  If your build is succeeding it is safe to ignore this warning."
```
Redth, line 100: Is this sane? Will this warning cause more panic than it's worth?

mattleibow: It looks to be fine. We do want a visible warning for the cases where `XamarinBuildDownloadRestoreAssemblyAar` is used - since we are no longer supporting that op.

mattleibow (approved):

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/Xamarin.Build.Download.targets
mfeingol: A question: I'm seeing this warning in my Xamarin.Forms project. However, I do not use `XamarinBuildDownloadRestoreAssemblyAar` anywhere. Seems like the warning may be overactive?

```1>C:\Users\mfeingol\.nuget\packages\xamarin.build.download\0.8.0\buildTransitive\Xamarin.Build.Download.targets(98,3): warning XBD100: ItemGroup `XamarinBuildDownloadRestoreAssemblyAar` is no longer supported.  If your build is failing, revert to an older Xamarin.Build.Download version, or migrate your packages to use the new `XamarinBuildDownloadAndroidAarLibrary` ItemGroup.  If your build is succeeding it is safe to ignore this warning.```

PR #794, opened 2020-02-13
Updated ARCore to 1.15.0

jparismorgan:
This change updates ARCore bindings to version 1.15.0 (based on #629). I have run the HelloAR sample app and verified that it works correctly. Let me know if I need to do any other validation.

Since the SceneForm projects depend on the ARCore NuGet package, I'll follow up with updates to the SceneForm bindings separately once the ARCore 1.15.0 NuGet package is available.

### DIFF
--- a/Android/ARCore/build.cake
+++ b/Android/ARCore/build.cake
@@ -2,9 +2,9 @@
 
 var TARGET = Argument ("t", Argument ("target", "Default"));
 
-var NUGET_VERSION = "1.12.0";
+var NUGET_VERSION = "1.15.0";
 
-var AAR_VERSION = "1.12.0";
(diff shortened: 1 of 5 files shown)


jparismorgan:
In doing other validation, I realized that ARFoundation doesn't support 1.15 yet, so we may instead be bumping to 1.14. If that's the case, I'll submit another PR moving to 1.14, then check this one in after (so we get both 1.14 and 1.15 for Xamarin). Would be create if we can hold off merging this until then.

This is the 1.14.0 PR #795 which this one should wait on.

https://docs.unity3d.com/Packages/com.unity.xr.arcore@3.1/changelog/CHANGELOG.html

Redth (approved):

PR #795, opened 2020-02-13
Updated ARCore to 1.14.0

jparismorgan:
This change updates ARCore bindings to version 1.14.0 (based on #629). I have run the HelloAR sample app and verified that it works correctly. Let me know if I need to do any other validation.

Since the Sceneform projects depend on the ARCore NuGet package, I'll follow up with updates to the Sceneform bindings separately once the ARCore 1.14.0 NuGet package is available.

This PR should go in before #794 - I'm doing both so that we can bump Azure Spatial Anchors to 1.14.0 now, and also have 1.15.0 ready (for when ARFoundation supports it).

### DIFF
--- a/Android/ARCore/build.cake
+++ b/Android/ARCore/build.cake
@@ -2,9 +2,9 @@
 
 var TARGET = Argument ("t", Argument ("target", "Default"));
 
-var NUGET_VERSION = "1.15.0";
+var NUGET_VERSION = "1.14.0";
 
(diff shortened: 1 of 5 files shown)


Redth (approved):

jparismorgan:
Hey @Redth, checking in if there's anything else I need to do to get this in? Doesn't look like I have permissions to merge.

PR #797, opened 2020-02-19
[XBD] Removes logic that embeds iOS resources using Cecil

SotoiGhost:
This also adds logic that cast old `RestoreAssemblyResource` tags into `BundleResource` tags to keep backwards compatibility on iOS.


mattleibow (approved):
Looking good!

PR #810, opened 2020-03-09
Fixes for Material Components Rc3

### LINKED ISSUE #807: Should class ContainerScheme implement IContainerScheming on the new Material v92
should `class ContainerScheme` implement `IContainerScheming`?

The iOS material docs seem to say I should be able to do this
```let scheme = MDCContainerScheme()
textField.applyTheme(withScheme: scheme)
```

But I'm not able to use `MDCContainerScheme` in this way because it doesn't implement `IContainerScheming` and ApplyTheme only takes `IContainerScheming`

I'm trying to figure out how to migrate this code
`TextFieldTypographyThemer.ApplyTypographyScheme(textField.TypographyScheme, textField.TextInput);`

it seems like it implements that interface inside the ApiDefinitions

https://github.com/xamarin/XamarinComponents/blob/269134acc384ad26b87fb12da35a91175a3b45c8/iOS/MaterialComponents/source/MaterialComponents/ApiDefinition.cs#L1200

But then in the code it doesn't seem like that interface is annotated on the `ContainerScheme` class
``` public class ContainerScheme : NSObject```

davehump:
Should fix #807


mattleibow (approved):
LGTM

PR #811, opened 2020-03-09
Guava 27.1.0.4 ReferenceJar -> EmbeddedReferenceJar

moljac:


Redth (approved):

mattleibow:
Why was this embedded? If this package is referenced, then the jar would also be referenced...

moljac:
Well making it `EmbeddedReferenceJar` solved several issues. 

More info here:

https://github.com/xamarin/AndroidX/issues/74

and here:

https://github.com/xamarin/GooglePlayServicesComponents/issues/308

I'm glad you joined discussion and review. If it only would be a bit earlier to help us out.

PR #822, opened 2020-03-30
OfficeUIFabric bindings 0.3.9. DONE

moljac:
Sample must be checked.


XPlat/OfficeUIFabric/Android/source/Additions/Additions.cs
```diff
@@ -17,30 +17,25 @@ public partial class Spacing
...
+                //        JniArgumentValue* __args = stackalloc JniArgumentValue[2];
+                //        __args[0] = new JniArgumentValue(cellPadding);
+                //        __args[1] = new JniArgumentValue(insetLeft);
+                //        var __rm = _members.InstanceMethods.InvokeNonvirtualObjectMethod(__id, this, __args);
+                //        return global::Java.Lang.Object.GetObject<global::Java.Interop.IJavaPeerable>(__rm.Handle, JniHandleOwnership.TransferLocalRef);
+                //    }
+                //    finally
+                //    {
+                //    }
+                //}
```
mattleibow, line 35: I would just delete the commented out code. Unless it has some significance?

```suggestion
```

mattleibow (approved):

XPlat/OfficeUIFabric/Android/source/Additions/Additions.cs
moljac: You would. I will.
You know me. I try to do as much as possible in metadata transforms and was not sure if that will work immediately, so I kept the source. (Not fond of reading through the git history)

moljac: Besides. I need to check that sample

PR #824, opened 2020-04-02
Xamarin.Kotlin.BindingSupport 0.7.0-preview

antonioseric:
update packages on native app to suport kotlin 1.3.71


mattleibow (approved):
Looks good! Thanks for the PR.

Just a note for going forward: we have deprecated this NuGet in favor of the binding project. Most, if not all, issues are fixed when building Kotlin bindings.

mattleibow:
Actually closing this as this is no longer supported and also, the core Android tooling is far superior.

We probably need to update the Kotlin bindings with the new tooling. Or maybe it has already, need to check.

PR #830, opened 2020-04-10
[Guava] Add minimal ProGuard configuration file

brendanzagaeski:
Context: https://github.com/xamarin/xamarin-android/issues/4369#issuecomment-610553571
Context: https://developercommunity.visualstudio.com/content/problem/979562/r8-warning-missing-class-javalangclassvalue.html
Context: https://github.com/xamarin/AndroidX/issues/74

Add a minimal ProGuard configuration file to stop the following R8
warning and error:

	R8 : warning : Missing class: java.lang.ClassValue
	R8 : error : Compilation can't be completed because some library classes are missing.

This warning appears because the Guava library is cross-platform.  The
library [determines at run time whether `ClassValue` is available or
whether to fall back to a different implementation][0].  On Android,
`android.jar` does *not* include `ClassValue`, so R8 warns about it at
build time, but the warning is safe to ignore because the library will
use the fallback.  In addition, the feature that depends on `ClassValue`
is [discouraged for use on Android][1] anyway:

	# Futures.getChecked (which often won't work with Proguard anyway) uses this. It
	# has a fallback, but again, don't use Futures.getChecked on Android regardless.
	-dontwarn java.lang.ClassValue

Example scenario that currently produces the warning:

 1. Create a Xamarin.Android bindings project that references the
    Xamarin.Google.Guava NuGet package.

 2. Create a Xamarin.Android app project that references the bindings
    project.

 3. Configure the app project to use multidex and the D8 DEX compiler so
    that it will use R8 for multidex.

 4. Attempt to package the app project, for example by running:

    	msbuild -restore -p:Configuration=Release -t:PackageForAndroid

[0]: https://github.com/google/guava/blob/9537f78cc652a4f5e77f4f418d5bbf4f49674515/guava/src/com/google/common/util/concurrent/FuturesGetChecked.java#L108
[1]: https://github.com/google/guava/wiki/UsingProGuardWithGuava/403775187dce3902c3186ed6153b011fbf69ab65


brendanzagaeski:
Ohhh shoot. This won't work with the way the Xamarin.Google.Guava NuGet package is intended to be used right now. The app project doesn't itself need to reference the Xamarin.Google.Guava NuGet package, so it won't pick up the `@(ProguardConfiguration)` item.

I'll think about this a little more.

brendanzagaeski:
*Changes in the latest commit:*

After thinking about the scenario a bit more, I realized that the
current way the Xamarin.Google.Guava NuGet package works is atypical.
Typically, when one NuGet package depends on another NuGet package, the
consuming NuGet package does not repackage the contents of the
dependency to be part of the new NuGet package.  Instead, the final
consuming application is expected to reference both NuGet packages,
either directly or transitively.

With the current approach, where the Xamarin.Google.Guava NuGet package
uses the `@(EmbeddedReferenceJar)` item type, the final consuming
application does *not* need to reference the Xamarin.Google.Guava NuGet
package at all, even transitively.  This breaks the usual conventions
for how NuGet packages behave.  That might cause some confusion for
bindings library authors.  It also makes it awkward to provide the
correct ProGuard configuration file to the final application
automatically.

Instead of including `guava.jar` as a `@(EmbeddedReferenceJar)`, revert
back to including it as a `@(ReferenceJar)`, but this time, *also*
include `guava.jar` as an `@(AndroidJavaLibrary)`.

When a bindings project references the NuGet package, the
`@(ReferenceJar)` item type will tell the build process to reference
`guava.jar`, allowing Java libraries that depend on it to be bound.  In
this scenario, the `@(AndroidJavaLibrary)` item type will be ignored.

When an application project references the NuGet package, the
`@(AndroidJavaLibrary)` item type will instead tell the build process to
package `guava.jar` as part of the application.  In this scenario, the
`@(ReferenceJar)` item type will be ignored.

Other changes:

Replace the `@(ProguardConfiguration)` item with a target that sets up
the configuration file by appending an argument to
`$(AndroidR8ExtraArguments)` instead.  This is required for now because
when R8 is configured for multidex only with no code shrinking, the
`@(ProguardConfiguration)` items are not used.

PR #831, opened 2020-04-12
[Guava] Change guava.jar to ReferenceJar + AndroidJavaLibrary

brendanzagaeski:
Context: https://github.com/xamarin/XamarinComponents/pull/830
Partially reverts: 362ed6a0e748bdd3288614b018bd62bdd7abab56
Context: https://github.com/xamarin/AndroidX/issues/74
Context: https://github.com/xamarin/GooglePlayServicesComponents/issues/308

While considering the best way to add a ProGuard configuration file to
the Xamarin.Google.Guava NuGet package, I realized that the NuGet
package was set up unusually.

Normally, when one NuGet package depends on another NuGet package, the
consuming NuGet package does not repackage the contents of the
dependency.  Instead, the final consuming application is expected to
reference both NuGet packages, either directly or transitively.

For the last couple of versions, the Xamarin.Google.Guava NuGet package
has been breaking this convention.  Those recent versions include a
`.targets` file that sets `guava.jar` to be an `@(EmbeddedReferenceJar)`
in any bindings library that references Xamarin.Google.Guava.  This
means that `guava.jar` is repackaged into the output assembly of the
bindings library.  As a result, any downstream app that consumes the
bindings library does not need to follow the normal convention of
referencing the Xamarin.Google.Guava NuGet package.  This seems likely
to cause confusion, and it also makes it awkward to include a ProGuard
configuration file in the NuGet package.

To resolve this, revert back to the approach from commit 14a3ada3 where
the `.targets` file was setting `guava.jar` to be a `@(ReferenceJar)`
instead of an `@(EmbeddedReferenceJar)`, but this time, also include
`guava.jar` as an `@(AndroidJavaLibrary)` for application projects.

When a bindings project references the NuGet package, the
`@(ReferenceJar)` item type tells the build process to reference
`guava.jar` but not repackage it into the output assembly.

When an application project references the NuGet package, the
`@(AndroidJavaLibrary)` item type tells the build process to package
`guava.jar` as part of the application.

I locally verified that the new `.targets` file worked correctly for
both of these scenarios.

The problem with the original approach in commit 14a3ada3 was that it
only included the `guava.jar` file as a `@(ReferenceJar)` item.  That
worked for bindings projects, but it didn't work for application
projects because application projects ignore the `@(ReferenceJar)` item
type.  One place this problem surfaced was in app builds using R8, as
seen in xamarin/AndroidX#74:

	R8 : warning : Missing class: com.google.common.util.concurrent.FutureCallback
	R8 : error : Compilation can't be completed because some library classes are missing.

Another place the problem surfaced was in running apps, as seen in
xamarin/GooglePlayServicesComponents#308:

	Java.Lang.NoClassDefFoundError: Failed resolution of: Lcom/google/common/base/Preconditions;

Admittedly, the only NuGet package maintained by the Xamarin team that
still references Xamarin.Google.Guava in the latest published version is
Xamarin.Firebase.Firestore, so there probably aren't too many apps that
need `guava.jar`, but it still seems like it would be nice to fix the
unconventional behavior so that Xamarin.Google.Guava can provide a
ProGuard configuration file for those apps.

Note that Xamarin.AndroidX.Room.Guava version 2.0.0 also referenced
Xamarin.Google.Guava, but the latest vesrion 2.2.5 does not because
androidx.room:room-guava [has switched][0] from Guava to the smaller
androix.concurrent:concurrent-futures library.  A likely motivation
behind that change is that the full Guava library is fairly large.

[0]: https://android.googlesource.com/platform/frameworks/support/+/2d074149f0430223ea9db69b268f6bcfeb1d4b4a%5E!

### DIFF
--- a/Android/Guava/source/Guava/Guava.targets
+++ b/Android/Guava/source/Guava/Guava.targets
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <EmbeddedReferenceJar Include="$(MSBuildThisFileDirectory)*.jar" />
+    <ReferenceJar Condition=" '$(AndroidApplication)' != 'true' " Include="$(MSBuildThisFileDirectory)*.jar" />
+    <AndroidJavaLibrary Condition=" '$(AndroidApplication)' == 'true' " Include="$(MSBuildThisFileDirectory)*.jar" />
   </ItemGroup>
 </Project>


brendanzagaeski:
Ooo, https://github.com/xamarin/GooglePlayServicesComponents/issues/317 is interesting. I'll see if I can look at that locally. The new approach in this PR of using `@(ReferenceJar)` combined with `@(AndroidJavaLibrary)` might actually help with that too.

For `Missing class: java.lang.ClassValue`, that warning is sort of ironically a *good* thing to see in builds because it means that `guava.jar` *is* being included in the application. (In particular, it means `guava.jar` is passed to `r8` when building the application.) As you mentioned, the correct fix is to use an R8 configuration file. I started working on that in https://github.com/xamarin/XamarinComponents/pull/830, but I haven't yet had a chance to update the PR to rebase it on this one (https://github.com/xamarin/XamarinComponents/pull/831). I'll do that soon.

In case it might also be of interest, I wrote up a user-facing explanation of what's happening with the `Missing class: java.lang.ClassValue` warning:
https://github.com/xamarin/xamarin-android/blob/4d8a78f4d094a53a2b0e8c320cbd932cd0ffd57e/Documentation/release-notes/4509.md#how-to-solve-missing-class-javalangclassvalue

jonpryor:
Let me start out by stating that I *don't* fully understand the entire ecosystem or relationships here.

That said, I don't understand this change.  Currently, `@(ReferenceJar)` and `@(EmbeddedReferenceJar)` are only supported in Bindings projects (those which import `Xamarin.Android.Bindings.targets)`, while `@(AndroidJavaLibrary)` only works in App/Library projects (those which import `Xamarin.Android.Common.targets`).  I thus fail to understand how replacing `@(ReferenceJar)` with `@(EmbeddedReferenceJar)`+`@(AndroidJavaLibrary)` would actually do anything.

I am thus confused.

Possibly (un?)related: https://github.com/xamarin/GooglePlayServicesComponents/issues/317#issuecomment-613526574

jonpryor:
> That said, I don't understand this change…

The NuGet package can be included by *both* Binding projects & App/Library projects.  Thus, it needs different item groups for the different referencing project types.

PR #863, opened 2020-05-14
Fix for silent Windows builds failuers because of CocoaPods

moljac:
Windows builds fail silently on DevOps and there are no nugets to sign. This prevents CocoaPods failure in Cake script

```
if ( ! IsRunningOnWindows() )
```

### DIFF
--- a/XPlat/Chromium.CroNet/build.cake
+++ b/XPlat/Chromium.CroNet/build.cake
@@ -65,20 +65,24 @@ Task ("externals")
 			}
 
 			EnsureDirectoryExists("./externals/ios");
-			if (CocoaPodVersion () < new System.Version (1, 0))
+
+			if ( ! IsRunningOnWindows() )
 			{
-				CocoaPods.RemoveAt (1);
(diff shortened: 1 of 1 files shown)


XPlat/Chromium.CroNet/build.cake
```diff
@@ -65,20 +65,24 @@ Task ("externals")
...
+				{
+					CocoaPods.RemoveAt (1);
+				}
 
-			FileWriteLines ("./externals/ios/Podfile", CocoaPods.ToArray ());
-			
-			CocoaPodRepoUpdate ();
+				FileWriteLines ("./externals/ios/Podfile", CocoaPods.ToArray ());
+				
+				CocoaPodRepoUpdate ();
```
grendello, line 78: Don't know if it matters for you, but indentation is different here?

grendello (approved):

XPlat/Chromium.CroNet/build.cake
moljac: Could be. I think I opened it in VS Code and I was working in VS2019. Not sure

PR #866, opened 2020-05-14
Added new version of Picasso

mattleibow:
Originally located here: https://github.com/mattleibow/square-bindings


tranb3r:
what about other square bindings ? are you gonna move them here as well ? are you still maintaining them in the original repo ?

PR #899, opened 2020-06-10
[ios][ExposureNotification] Update bindings to Xcode 11.6 Beta 1

dalexsoto:


iOS/ExposureNotification/source/ExposureNotification.csproj
```diff
@@ -5,8 +5,8 @@
     <IsBindingProject>true</IsBindingProject>
     <RootNamespace>ExposureNotification</RootNamespace>
     <AssemblyName>Xamarin.iOS.ExposureNotification</AssemblyName>
-    <AssemblyVersion>1.0.0.0</AssemblyVersion>
-    <FileVersion>1.0.0.0</FileVersion>
+    <AssemblyVersion>1.1.0.0</AssemblyVersion>
+    <FileVersion>1.1.0.0</FileVersion>
```
mattleibow: Do we want to do this? Do we not keep the actual assembly version at 1.0.0.0 and just bump the information part?

iOS/ExposureNotification/source/ApiDefinition.cs
```diff
@@ -106,6 +114,15 @@ interface ENExposureConfiguration {
 		[Export ("minimumRiskScore")]
 		byte MinimumRiskScore { get; set; }
 
+		[Introduced (PlatformName.iOS, 13, 6)]
+		[Export ("minimumRiskScoreFullRange")]
+		double MinimumRiskScoreFullRange { get; set; }
+
+		[Introduced (PlatformName.iOS, 13, 6)]
+		[BindAs (typeof (int []))]
```
mattleibow, line 122: What does the `[BindAs (typeof (int []))]` actually do when the return type is `NSNumber[]`?

iOS/ExposureNotification/source/ExposureNotification.csproj
dalexsoto:
```suggestion
    <AssemblyVersion>1.0.0.0</AssemblyVersion>
    <FileVersion>1.0.0.0</FileVersion>
```

iOS/ExposureNotification/source/ApiDefinition.cs
dalexsoto: This means that we will take care of the conversion of `NSNumber []` <-> `int []` automatically in the binding code so the signature will become

```csharp
int [] AttenuationDurationThresholds { get; }
```

So you do not need to worry about the NSNumber [] detail

mattleibow: That is very nice.

mattleibow (approved):

PR #905, opened 2020-06-25
Add Xamarin.Firebase.Messaging as Notification Hubs Dependency

marstr:
PR 1 of 2

It's come to our attention that the existing Android Notification Hub sample is inoperable. The trouble is a missing Xamarin wrapper of the `FirebaseReceiver` type. The fix is fairly simple, add the dependency to Xamarin.Firebase.Messaging necessary for the Receiver to be discovered and emitted when building the binding library.

To fully remedy the situation, we'll also need to update the Sample Application to target the new binary. I've confirmed locally that once these changes are made, that we can initialize Firebase and receive notifications to the device. However, because the sample takes a dependency on the NuGet version of this package, we must break this into two PRs.

### DIFF
--- a/XPlat/AzureMessaging/Android/source/Xamarin.Azure.NotificationHubs.Android.csproj
+++ b/XPlat/AzureMessaging/Android/source/Xamarin.Azure.NotificationHubs.Android.csproj
@@ -42,6 +42,7 @@
 
   <ItemGroup>
     <PackageReference Include="Xamarin.Android.Support.v7.AppCompat" Version="28.0.0.3" PrivateAssets="None" />
+    <PackageReference Include="Xamarin.Firebase.Messaging" Version="71.1740.1" />
   </ItemGroup>
 
 </Project>
--- a/XPlat/AzureMessaging/Android/source/Xamarin.Azure.NotificationHubs.Android.csproj
+++ b/XPlat/AzureMessaging/Android/source/Xamarin.Azure.NotificationHubs.Android.csproj
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="MSBuild.Sdk.Extras/2.0.54">
   <PropertyGroup>
-    <TargetFramework>MonoAndroid80</TargetFramework>
+    <TargetFramework>MonoAndroid90</TargetFramework>
     <IsBindingProject>true</IsBindingProject>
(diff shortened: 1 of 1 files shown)


mattleibow (approved):

mattleibow:
I had a look at the diff and I see this:

#### Type Changed: WindowsAzure.Messaging.Registration.RegistrationType

Obsoleted properties:

```diff
 [Obsolete ("deprecated")]
 public static Registration.RegistrationType Gcm { get; }
```

#### New Type: WindowsAzure.Messaging.NotificationHubs.FirebaseReceiver

```csharp
public sealed class FirebaseReceiver : Firebase.Messaging.FirebaseMessagingService {
	// constructors
	public FirebaseReceiver ();
	public FirebaseReceiver (NotificationHub hub);
	// properties
	public override Java.Interop.JniPeerMembers JniPeerMembers { get; }
	protected override IntPtr ThresholdClass { get; }
	protected override System.Type ThresholdType { get; }
}
```

There are no properties on this, just the ctor. Is this correct?

marstr:
No properties, should be a few methods though. Do you see `onCreate`, `onMessageReceived`, `onNewToken`,  (static) `getNotificationMessage`?

mattleibow:
Is that maybe on the base type? This diff is generated, so it should be everything.

mattleibow:
Seems to be on the base type: https://www.fuget.org/packages/Xamarin.Firebase.Messaging/71.1740.1/lib/monoandroid90/Xamarin.Firebase.Messaging.dll/Firebase.Messaging/FirebaseMessagingService

mattleibow:
Maybe I merged too fast... I didn't see `getNotificationMessage`... And it appears to not be a Firebase things but NH?

mattleibow:
The static method appears to be private... (Java code)

```java
public final class FirebaseReceiver extends FirebaseMessagingService {
  public FirebaseReceiver();
  public FirebaseReceiver(NotificationHub hub);
  public void onCreate();
  public void onMessageReceived(@NonNull RemoteMessage remoteMessage);
  public void onNewToken(@NonNull String s);
  static BasicNotificationMessage getNotificationMessage(RemoteMessage remoteMessage);
}
```

Seems that `BasicNotificationMessage` is also not public...

marstr:
No no, this seems alright. `getNotificationMessage` should be private, as should `BasicNotificationMessage`.

Thanks!

PR #907, opened 2020-06-26
Add nh firebase dependency

marstr:
Waded through the dependency tree, made the change necessary to have a minimum set of dependencies needed to run this application. After this, all a person should have to do is register their app with Firebase, download the `google-services.json` and add it to the Properties portion of their application and set the Build action to 'GoogleServicesJson'.

### DIFF
--- a/XPlat/AzureMessaging/.gitignore
+++ b/XPlat/AzureMessaging/.gitignore
@@ -1,0 +1,1 @@
+google-services.json
--- a/XPlat/AzureMessaging/Android/samples/NotificationHubsSample/NotificationHubsSample.csproj
+++ b/XPlat/AzureMessaging/Android/samples/NotificationHubsSample/NotificationHubsSample.csproj
@@ -54,6 +54,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <GoogleServicesJson Include="Properties\google-services.json" />
     <None Include="Properties\AndroidManifest.xml" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xamarin/XamarinComponents?pullRequest=907) <br/>All CLA requirements met.

marstr:
I guess one notable change, is that the build process now directly targets the Project, instead of creating a Local Nuget, and having the build phase publish there.

Now that I understand better what is happening, I could work on altering it back to that. Personally I think it's a bit of unnecessary complication, but I'll yield to other folk's opinions on this. The big pro of altering this back would be that it more closely resembles an outside project's acquisition path.

mattleibow (approved):

mattleibow:
In most cases, we just use project references. This project might have used NuGets for some reason - maybe @Redth will have an idea. But, I think this is just nicer for general dev.

marstr:
Nice, I'm happy with this then. Once this is merged, we can release an updated version of the NuGet as well (the existing one doesn't work.) Maybe call it `v1.0.0-preview2.1`?

mattleibow:
@marstr we can do that!

When updating the packages, just update the version number in the build.cake file. That is used when creating the package. I did that for now, and when CI is all green, I'll release.

Thanks for this!

mattleibow:
@marstr it is LIVE!!! 

https://www.nuget.org/packages/Xamarin.Azure.NotificationHubs.Android/1.0.0-preview2.1

marstr:
Thanks @mattleibow!

PR #937, opened 2020-08-04
Play.Core-1.7.3

moljac:
Play.Core-1.7.3 version bumps only


Android/Google.Play.Core/source/Xamarin.Google.Play.Core.XamarinAndroid/Xamarin.Google.Play.Core.XamarinAndroid.csproj
```diff
@@ -8,7 +8,7 @@
     <AndroidUseIntermediateDesignerFile>True</AndroidUseIntermediateDesignerFile>
     <AndroidResgenFile>Resources\Resource.designer.cs</AndroidResgenFile>
 
-    <TargetFramework>MonoAndroid90</TargetFramework>
```
jpobst: Is there a reason to multitarget this?

azure-pipelines.yml
```diff
@@ -43,13 +43,13 @@ jobs:
             dotnet tool uninstall --global Cake.Tool
             dotnet tool install --global Cake.Tool
             dotnet tool install --global boots
-            boots https://aka.ms/xamarin-android-commercial-d16-8-macos
```
jpobst: Is this required?  This is going to affect building *all* packages in this repository, and there's no guarantee that switching to a preview version isn't going to break other packages or cause API to disappear because it is no longer bound.

Android/Google.Play.Core/source/Xamarin.Google.Play.Core.XamarinAndroid/Xamarin.Google.Play.Core.XamarinAndroid.csproj
moljac: No. Tests.

azure-pipelines.yml
moljac: No. It is not required.

Version 1.6.x was built w/o any problems and just bump to 1.7.x caused weird build errors, so I tried "updating the tooling". namely it worked locally - as usual.
Removing that

jpobst (approved):

PR #938, opened 2020-08-04
Play.Core-1.8.0

moljac:
Play.Core-1.8.0 version bumps only


uwe-neuronation:
Any progress here? Play.Core is an imported part of a modern Android App.
Would be nice if someone could take care.
please

jpobst (approved):

PR #940, opened 2020-08-05
Xamarin AzureMessaging GA for new APIs

mpodwysocki:
This PR will update the AzureMessaging Xamarin libraries to 3.0 for Apple and 1.0 for Android.  Apple support now includes tvOS, macOS and iOS.


XPlat/AzureMessaging/iOS/source/ApiDefinition.cs
```diff
@@ -166,6 +167,9 @@ public partial interface MSNotificationHub
 		[Static, Export("startWithConnectionString:hubName:")]
 		void Start(string connectionString, string hubName);
 
+		[Static, Export("StartWithInstallationManagement:")]
+		void StartWithInstallationManagement(MSInstallationManagementDelegate managementDelegate);
```
mattleibow: It might just be better to call this `Start` and have it as a simple overload?

mpodwysocki: Renamed to start

mattleibow:
Seems CI is giving a 404 when downloading the android bits?

mpodwysocki:
> Seems CI is giving a 404 when downloading the android bits?

Correct @mattleibow, that's why it's in draft right now because we're waiting on the Android release.  Once it has been released, I'll move it to active.

mpodwysocki:
@mattleibow not sure why CI seems to be failing, but doesn't seem related to any changes I've made

mattleibow (approved):

PR #963, opened 2020-08-29
Retrofit2 2.8.2

antonioseric:


mattleibow:
I am going to add OkHttp3 v3 back as a separate build here. I think that version is still updating. v4 is Kotlin based and this might not be what people want to do just yet.

antonioseric:
@mattleibow @moljac this version is not released to nuget.org

PR #968, opened 2020-09-04
Add back the v3 version of OkHttp3

mattleibow:
The v4 is for Kotlin, and v3 seems to still gets updates often


antonioseric:
@mattleibow @moljac OkHttp3 3.14.9 is not released on nuget.org

PR #986, opened 2020-09-16
OkHttp3-4.3.1

antonioseric:


tranb3r:
When is the nuget going to be released ?

4brunu:
I think it's not published yet.

tranb3r:
What is the process to get the nuget published once the PR is merged ?

4brunu:
I don't know, maybe @mattleibow can help with that.

mattleibow:
I think this was a typical case of me merging and then going to bed. If all goes well, people get alerts and they hit the button. But, seems a step failed, so nothing got sent out....
Anyway, did a rebuild and now they are out in the wild!!! Go moggy!

Issue #966, opened 2020-09-02
Update ConstraintLayout package


petro2050:
I just ran into this problem: https://stackoverflow.com/questions/57251887/android-motionlayout-android-resource-linking-failed after updating VS to its latest. My project doesn't compile anymore.

Can you update the `ConstraintLayout` package to the latest version?

moljac:
published

PR #989, opened 2020-09-19
ConstraintLayout 2.0.1

antonioseric:
latest version for ConstraintLayout #966 .

@mattleibow we may have issue with this binding, not sure how to resolve it. 
In `MotionLayout` there are 2 Methods (`AddTransitionListener` and `SetTransitionListener`) using interface `ITransitionListener` 
and in region Event implementation for Android.Support.Constraints.Motion.MotionLayout.ITransitionListener there are two implemenatation of `ITransitionListener` interface so we have error that MotionLayout already contains definition for interface methods.... 
We can remove one of methods  (`AddTransitionListener` and `SetTransitionListener`) and it will work but that is not good solution. Do you have idea how to fix it? or it is maybe Xamarin.Android bug? 

I also find simular issue https://forums.xamarin.com/discussion/85194/android-binding-library-question


mattleibow:
So the add/remove are the better options to keep, so we can do things to the set.

I believe there is an "eventName" thing you can set to a blank string that might prevent the event generation.

```xml
<attr path="xxx" name="eventName"></attr>
```

That might do it.

Android/ConstraintLayout/source/ConstraintLayoutSolver/Transforms/Metadata.xml
```diff
@@ -13,5 +13,13 @@
...
 	<attr path="/api/package[@name='android.support.constraint.solver.widgets']/class[@name='Guideline']/field[@name='VERTICAL']" name="managedName">VERTICAL</attr>
 	<attr path="/api/package[@name='android.support.constraint.solver.widgets']/class[@name='Guideline']/field[@name='HORIZONTAL']" name="managedName">HORIZONTAL</attr>
+    
+    <attr path="/api/package[@name='android.support.constraint.solver.state']/class[@name='Dimension']/method[@name='Fixed' and count(parameter)=1 and parameter[1][@type='int']]" name="managedName">FixedFromValue</attr>
+    <attr path="/api/package[@name='android.support.constraint.solver.state']/class[@name='Dimension']/method[@name='Suggested' and count(parameter)=1 and parameter[1][@type='int']]" name="managedName">SuggestedFromValue</attr>
+    <remove-node path="/api/package[@name='android.support.constraint.solver.state']/class[@name='Dimension']/method[@name='Percent' and count(parameter)=2 and parameter[1][@type='java.lang.Object'] and parameter[2][@type='float']]" />
```
moljac, line 19: Why `remove-node`? No other way to bind?
just checking

moljac (approved):

PR #1001, opened 2020-10-22
Update signing to use the new logic

mattleibow:


mattleibow:
Waiting on the internal PR: https://github.com/xamarin/yaml-templates/pull/73

PR #1015, opened 2020-11-11
Update Xamarin iOS and Xamarin Android for AzureMessaging

mpodwysocki:
Updates iOS to 3.1.1 and Android to 1.1.2.


XPlat/AzureMessaging/iOS/source/ApiDefinition.cs
```diff
@@ -284,7 +284,7 @@ public partial interface MSNotificationHubMessage
 		string Body { get; }
 
 		[Export("userInfo")]
-		NSDictionary<NSString, NSString> UserInfo { get; }
+		NSDictionary UserInfo { get; }
```
mattleibow, line 287: This is a breaking change. Just need confirmation.
Also, the iOS version was not increased, so this will not actually leave the door.

XPlat/AzureMessaging/Android/source/Xamarin.Azure.NotificationHubs.Android.csproj
```diff
@@ -40,10 +41,20 @@
      <TransformFile Include="Transforms\*.xml" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('MonoAndroid90')) ">
     <PackageReference Include="Xamarin.Android.Support.v7.AppCompat" Version="28.0.0.3" PrivateAssets="None" />
-    <PackageReference Include="Xamarin.Firebase.Messaging" Version="71.1740.4" />
+    <PackageReference Include="Xamarin.Google.Android.DataTransport.TransportBackendCct" Version="2.2.2" />
+    <PackageReference Include="Xamarin.Google.Android.DataTransport.TransportRuntime" Version="2.2.2" />
+    <PackageReference Include="Xamarin.Firebase.Messaging" Version="120.1.7" />
```
mattleibow: This version depends on AndroidX. I don't think we need to use Android Support anymore. All the libraries used here are AndroidX, so we can drop 90 and make this library AndroidX only.
If the user is still using Android Support, then the upgrade process will migrate everything.

mattleibow:
I think this is related as well: https://github.com/xamarin/XamarinComponents/issues/1016

XPlat/AzureMessaging/iOS/source/ApiDefinition.cs
mpodwysocki: Yes, this is a breaking change due to a miscommunication on our team.  This should have always been an `NSDictonary *` instead of` NSDictionary<NSString *, NSString *> *`. Currently, this doesn't work, because it casts an `InvalidCastException` due to this bug

mpodwysocki: Will update to 3.1.1 then, since this was an issue.

XPlat/AzureMessaging/Android/source/Xamarin.Azure.NotificationHubs.Android.csproj
mpodwysocki: Ok, removed all old compat support

mpodwysocki:
@mattleibow Correct, it is related. As noted in https://github.com/Azure/azure-notificationhubs-xamarin/issues/13 also, that this is a breaking change, but needs to be made since the existing code is broken as is.

XPlat/AzureMessaging/build.cake
```diff
@@ -1,12 +1,12 @@
...
+var IOS_VERSION = "3.1.1";
+var IOS_NUGET_VERSION = "3.1.1";
 var IOS_URL = $"https://github.com/Azure/azure-notificationhubs-ios/releases/download/{IOS_VERSION}/WindowsAzureMessaging-SDK-Apple-{IOS_VERSION}.zip";
 
-var ANDROID_VERSION = "1.1.1";
-var ANDROID_NUGET_VERSION = "1.1.1";
-var ANDROID_URL = string.Format ("https://dl.bintray.com/microsoftazuremobile/SDK/com/microsoft/azure/notification-hubs-android-sdk/{0}/notification-hubs-android-sdk-{0}.aar", ANDROID_VERSION);
+var ANDROID_VERSION = "1.1.3";
+var ANDROID_NUGET_VERSION = "1.1.3";
+var ANDROID_URL = $"https://dl.bintray.com/microsoftazuremobile/SDK/com/microsoft/azure/notification-hubs-android-sdk/{0}/notification-hubs-android-sdk-{ANDROID_VERSION}.aar";
```
mattleibow:
```suggestion
var ANDROID_URL = $"https://dl.bintray.com/microsoftazuremobile/SDK/com/microsoft/azure/notification-hubs-android-sdk/{ANDROID_VERSION}/notification-hubs-android-sdk-{ANDROID_VERSION}.aar";
```

mattleibow (approved):

PR #1038, opened 2020-11-26
Install xharness onto the bots

mattleibow:


premun:
@mattleibow maybe not the case here but be aware that if you call `dotnet tool install` and the tool is already installed, the command returns non-zero exit code and it fails.

What we actually do to not clog our agents with many versions of XHarness, we install it on the build agent locally (in a working folder) and package it as payload that gets sent to Helix with the job. Then, the Helix agent doesn't even need .NET SDK but only .NET runtime and downloads this extracted package.
You could also install it locally before every job and then remove it with rest of the build files.

I can go into further details if you want and explain what we do.

mattleibow:
Thanks for the heads up! This template is mainly for running on the default Azure bots that do not have any tools installed. Once we start using shared bots then we can't use this template as is since it assumes many things. When we start using helix, then we probably won't even need the template as then all the work is done on those bots instead.

PR #1043, opened 2020-12-04
Use the microbuild version of signing

cadsit:
We're trying to remove all dependencies on the existing code-sign instance, in particular the `sign-from-vsts` job.

This is the last place that uses the pipeline, and I'm curious if we could just switch over to microbuild here instead.


mattleibow (approved):

mattleibow:
Just a note to all future builds, the output .dll, .jar and .aar files need to be added to the root SignList.xml

PR #1054, opened 2020-12-15
Square prefix for signing. Was: Update SignList.xml

moljac:


jpobst (approved):
Looks good to me, but I don't really know anything about this.  cc'ing @mattleibow as an FYI.

moljac:
@mattleibow I will collect all prefixes (or most of them). I hope 2021-01. When I'll be off the grid (vacation)

Issue #1066, opened 2021-01-04
ExposureNotification iOS 12.5 support


pavshr:
Hello, 

Recently, Apple have added support of [Exposure Notification to iOS 12.5](https://developer.apple.com/documentation/exposurenotification/supporting_exposure_notifications_in_ios_12_5). 

Is there any plans to add support for it in Xamarin.iOS and XPlat cross-plarform lib and can you provide us with any timeline? Thank you.

mattleibow:
Support for iOS 12.5 looks pretty simple at this stage. There are a few things that need to be done but I will get on it right away:
 - [ ] most code checks are for 13.5+, so these will need to change a bit
 - [ ] support for the special callback needs to be implemented

It is important to note that iOS 13.0-13.4 still won't get support for EN.

> Although iOS 13.0 to 13.4 doesn’t support Exposure Notifications, your app might still download onto devices with those versions of iOS. If that happens, present a dialog to let the user know Exposure Notifications isn’t available for their device.

I need to confirm, but it appears that only v2 notifications are available on iOS 12.5

> If you have an app that uses version 1 risk scoring and want to support iOS 12.5, consider updating to the more robust version 2 scoring.

mattleibow:
Work started here: https://github.com/xamarin/XamarinComponents/pull/1072

mattleibow:
I have just pushed the iOS NuGet out: https://www.nuget.org/packages/Xamarin.iOS.ExposureNotification/1.2.2-preview1

The cross-platform one is now building and I'll push that out soon.

mattleibow:
Both PRs have been merged. The iOS version is out and the xplat API is building and will go out as soon as it completes.

PR #1072, opened 2021-01-11
[EN] [iOS] Add support for iOS 12.5

mattleibow:
Apple backported some features to the 12.5 update: https://developer.apple.com/documentation/exposurenotification/supporting_exposure_notifications_in_ios_12_5

Usage of this should really be a check for "if iOS 12.5, then set the `LaunchActivityHandler` property, otherwise register a background task. 

```csharp
private ENManager GetManager ()
{
    // the type is not available
    if (ObjCRuntime.Class.GetHandle ("ENManager") == IntPtr.Zero)
        return null;

    // create the manager
    var manager = new ENManager ();

    if (DeviceInfo.Version < new Version (13, 5)) { 
        // backwards compat
        manager.SetLaunchActivityHandler (activityFlags => {
            if (activityFlags.HasFlag(ENActivityFlags.PeriodicRun)) {
                // Your app now has 3.5 minutes to perform download and detection.
            }
        });
    } else {
        // schedule background task using BGTaskScheduler
    }

    // ...

    return manager;
}
```

### DIFF
--- a/iOS/ExposureNotification/source/ApiDefinition.cs
+++ b/iOS/ExposureNotification/source/ApiDefinition.cs
@@ -357,6 +357,11 @@ namespace ExposureNotifications {
 	delegate void ENGetExposureWindowsHandler ([NullAllowed] ENExposureWindow [] exposureWindows, [NullAllowed] NSError error);
 	delegate void ENGetUserTraveledHandler (bool traveled, NSError error);
 
+	[Introduced (PlatformName.iOS, 12, 5)]
+	public enum ENActivityFlags : uint {
+		ENActivityFlagsPeriodicRun = 1U << 2
+	}
+
 	[Introduced (PlatformName.iOS, 13, 5)]
 	[BaseType (typeof (NSObject))]
 	interface ENManager {
@@ -369,6 +374,9 @@ namespace ExposureNotifications {
 
 		[NullAllowed, Export("invalidationHandler", ArgumentSemantic.Copy)]
 		Action InvalidationHandler { get; set; }
+
+		[NullAllowed, Export ("activityHandler", ArgumentSemantic.Copy)]
+		Action<ENActivityFlags> LaunchActivityHandler { get; set; }
 
 		[Async]
 		[Export ("activateWithCompletionHandler:")]
--- a/iOS/ExposureNotification/source/ApiDefinition.cs
+++ b/iOS/ExposureNotification/source/ApiDefinition.cs
@@ -357,11 +357,6 @@ namespace ExposureNotifications {
 	delegate void ENGetExposureWindowsHandler ([NullAllowed] ENExposureWindow [] exposureWindows, [NullAllowed] NSError error);
 	delegate void ENGetUserTraveledHandler (bool traveled, NSError error);
 
-	[Introduced (PlatformName.iOS, 12, 5)]
-	public enum ENActivityFlags : uint {
-		ENActivityFlagsPeriodicRun = 1U << 2
-	}
-
 	[Introduced (PlatformName.iOS, 13, 5)]
 	[BaseType (typeof (NSObject))]
 	interface ENManager {
@@ -374,9 +369,6 @@ namespace ExposureNotifications {
 
 		[NullAllowed, Export("invalidationHandler", ArgumentSemantic.Copy)]
 		Action InvalidationHandler { get; set; }
-
-		[NullAllowed, Export ("activityHandler", ArgumentSemantic.Copy)]
-		Action<ENActivityFlags> LaunchActivityHandler { get; set; }
 
 		[Async]
 		[Export ("activateWithCompletionHandler:")]
--- a/iOS/ExposureNotification/source/ExposureNotification.csproj
+++ b/iOS/ExposureNotification/source/ExposureNotification.csproj
@@ -22,7 +22,7 @@
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=2128732</PackageProjectUrl>
     <PackageLicenseFile>License.md</PackageLicenseFile>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageVersion>1.2.1</PackageVersion>
+    <PackageVersion>1.2.2</PackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/iOS/ExposureNotification/source/ExposureNotification.csproj
+++ b/iOS/ExposureNotification/source/ExposureNotification.csproj
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="MSBuild.Sdk.Extras/2.0.54">
   <PropertyGroup>
     <TargetFrameworks>Xamarin.iOS10</TargetFrameworks>
@@ -43,5 +42,4 @@
     <ObjcBindingApiDefinition Include="ApiDefinition.cs" Condition=" '$(EnableDefaultCompileItems)' == 'true'" />
   </ItemGroup>
 
-  <Import Project="$(MSBuildSDKExtrasTargets)" Condition="Exists('$(MSBuildSDKExtrasTargets)')" />
 </Project>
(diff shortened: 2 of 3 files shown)


iOS/ExposureNotification/source/ApiDefinition.cs
```diff
@@ -367,17 +352,19 @@ interface ENManager {
 		[Export ("exposureNotificationStatus", ArgumentSemantic.Assign)]
 		ENStatus ExposureNotificationStatus { get; }
 
-		[NullAllowed, Export("invalidationHandler", ArgumentSemantic.Copy)]
+		[NullAllowed, Export ("invalidationHandler", ArgumentSemantic.Copy)]
 		Action InvalidationHandler { get; set; }
 
+		[NullAllowed, Export ("activityHandler", ArgumentSemantic.Copy)]
+		Action<ENActivityFlags> LaunchActivityHandler { get; set; }
```
mattleibow: @dalexsoto is this the equivalent code for setting "value for key"? I know that there is a `SetValueForKey`, but I was hoping to leverage the binding engine...

```objc
typedef NS_OPTIONS( uint32_t, ENActivityFlags )
{
    /// The app launched to perform periodic operations.
    ENActivityFlagsPeriodicRun = ( 1U << 2 )
};

typedef void (^ENActivityHandler)(ENActivityFlags activityFlags);

@interface ENManager (ENActivityHandling)

@end

@implementation ENManager (ENActivityHandling)

- (void)setLaunchActivityHandler:(ENActivityHandler)activityHandler {
    [self setValue:activityHandler forKey:@"activityHandler"];
}

@end
```

iOS/ExposureNotification/source/ApiDefinition.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace ExposureNotifications {
 
-	[Introduced (PlatformName.iOS, 13, 5)]
+	[Introduced (PlatformName.iOS, 12, 5)]
```
mattleibow: Just checking if this will cause any issues.

What happens now is that iOS 12.5 has this feature (but not 13.0 - 13.4). It comes back in iOS 13.5.

spouliot: it will not cause runtime issue, the attributes are actually removed by the linker
but it can be confusing to developers, inside the IDE, as it means _since 12.5_

iOS/ExposureNotification/source/ApiDefinition.cs
```diff
@@ -59,7 +59,7 @@ public enum ENStatus : long {
 		Unauthorized = 6,
 	}
 
-	[Introduced (PlatformName.iOS, 13, 7)]
+	[Introduced (PlatformName.iOS, 12, 5)]
```
spouliot: that might be even mode confusing in this case, as 13.5/6 won't have it either

spouliot (changes requested):
I think you're better off not updating the attributes. It will likely cause more confusion than help.

You could add an `[Advice ("API also available on iOS 12.5+")]` (on each type) if you think it's useful - or just add this piece of information in the documentation / `README.md`

iOS/ExposureNotification/source/ApiDefinition.cs
spouliot: if the implementation only in the headers ?
in that case it might not be in the binary so you can't bind it
but you can provide a (manually written) method that does the same
you might be able to use `[Wrap("...")]` but that does not debug well and is not verified by the compiler (well it is after code is generated)

iOS/ExposureNotification/source/ApiDefinition.cs
mattleibow: Good point. Let me revert these.

iOS/ExposureNotification/source/ApiDefinition.cs
mattleibow: How does one properly wrap a `Action<T>` to `NSObject`? Is there a clean wrapper?

The sample usage was written like this:

```objc
// Set the launch activity handler.
[manager setLaunchActivityHandler:^(ENActivityFlags activityFlags) {
    if (activityFlags & ENActivityFlagsPeriodicRun) {
        // Your app now has 3.5 minutes to perform download and detection.
    }
}];
```

mattleibow: Also, the implementation is recommended to inject it yourself. Seems this value is something that only the older implementations look at. I think you are literally supposed to just set this from outside, but you wrap it in the "extensions" for niceness

```objc
[manager setValue:activityHandler forKey:@"activityHandler"];
```

dalexsoto: This may require you to create the obj-c block from a managed delegate yourself kinda like show in this file 

https://github.com/xamarin/xamarin-macios/blob/6a6f1f1ff8256e154acca6519d8bc56553e25845/src/Network/NWFramer.cs#L129-L147

But instead of passing the block pointer to a c function as shown above you may need to add this member to `ENManager` and give the block pointer to that function

```csharp
[Sealed]
[Export ("setValue:forKey:")]
void SetValueForKey (IntPtr ptr, NSString key);
```

This all depends on how/who calls into the the `activityHandler` block, if it is the native code then you have no other choice than doing this manually, if it is you from your C# code then you can avoid all of this and stay in managed land.

@spouliot do you agree? I hope I am not posting non-sense at 6 pm :P

spouliot: @dalexsoto agreed, with some ☕ to back it up 😄

iOS/ExposureNotification/source/ExposureNotification.csproj
```diff
@@ -43,5 +42,4 @@
     <ObjcBindingApiDefinition Include="ApiDefinition.cs" Condition=" '$(EnableDefaultCompileItems)' == 'true'" />
   </ItemGroup>
 
-  <Import Project="$(MSBuildSDKExtrasTargets)" Condition="Exists('$(MSBuildSDKExtrasTargets)')" />
```
dalexsoto, line 46: Is this meant to be done? is it not needed anymore?

dalexsoto (approved):
heh what a nice **block** of code

iOS/ExposureNotification/source/ExposureNotification.csproj
mattleibow: Not any more. The new Sdk attribute at the top does it for us :)

iOS/ExposureNotification/source/Extensions.cs
```diff
@@ -23,4 +26,42 @@ public static class ENRiskWeightRange {
 		public static byte Min { get; } = byte.MinValue;
 		public static byte Max { get; } = 100;
 	}
+
+	[Introduced (PlatformName.iOS, 12, 5)]
```
spouliot: I'd remove that one too

iOS/ExposureNotification/source/Extensions.cs
```diff
@@ -23,4 +26,42 @@ public static class ENRiskWeightRange {
...
+		private static unsafe void ENActivityHandlerCallback (IntPtr block, ENActivityFlags activityFlags)
+		{
+			var descriptor = (BlockLiteral*) block;
+			var del = (ENActivityHandler) (descriptor->Target);
+			del?.Invoke (activityFlags);
+		}
+
+		private static ENActivityHandlerCallbackDelegate ENActivityHandlerCallbackReference = ENActivityHandlerCallback;
+
+		public void SetLaunchActivityHandler (ENActivityHandler activityHandler)
```
spouliot: double check (test) with a `null` value - it might work (or throw)

If it throws then you might want to set `ptr = IntPtr.Zero;` instead of the `Block*` related code

mattleibow: wow... 🤦

mattleibow: Should I be running the CleanupBlock as well? Will that dispose anything that needs to stick around?

mattleibow: Also, I see some code in xamarin-macios that does not do a _Block_copy. Should I be doing a copy?

mattleibow:
This is the iOS changes for #1066

PR #1074, opened 2021-01-12
[EN] [XPlat] Add support for iOS 12.5

mattleibow:
**iOS 12.5 Support**

This PR adds the support for iOS 12.5 notifications. To enable this platform, no actual code needs to change. This is done by leveraging the `ScheduleFetch` method to determine which task registration method to use - either BGTask or the new way. This method is actually called by `Init`, so it happens early on.

**IsSupported**

On platforms that do not support EN, such as iOS 10 or iOS 13.2, then it will now throw. To detect these scenarios, there is a new property `IsSupported` that can be used to display messages to the user. If it is `true`, then it does not guarantee that everything is ready to be used, but if it is `false`, then it is guaranteed that EN _cannot_ be used. 

The reason for this is that it will typically return `true`, and then if the user denies permissions or (on Android) GPS is out of date, or on iOS there is some extra criteria. These will be handled by the normal exceptions (unchanged). When this returns false, it means that there is _no_ EN support at all. Either no APIs or GPS explicitly indicated that there is no EN support on the device. 

If there is a case where this property is returning `true` and it can be pre-determined that it should return false, feel free to open an issue and I will try make sure we handle those extra cases.

### DIFF
--- a/XPlat/ExposureNotification/build.cake
+++ b/XPlat/ExposureNotification/build.cake
@@ -1,7 +1,6 @@
 var TARGET = Argument("t", Argument("target", "ci"));
 
 var OUTPUT_PATH = (DirectoryPath)"./output/";
-var NUGET_VERSION = "0.13.0-preview";
 
 Task("nuget")
 	.Does(() =>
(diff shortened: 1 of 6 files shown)


mattleibow:
This is the xplat changes for #1066 and requires #1072

PR #1081, opened 2021-01-25
[azdo] make dotnet parameter optional

jonathanpeppers:
Context: https://github.com/xamarin/AndroidX/pull/247
Context: https://github.com/xamarin/Xamarin.Legacy.Sdk

I'm testing out if we can multi-target the AndroidX libraries for `MonoAndroid90;net6.0-android`.

Unfortunately the build fails because:

1. The `UseDotNet@2` yaml task can't install .NET
   6.0.100-alpha.1.21064.27. We have to use the `dotnet-install`
   script to install it ourselves.

2. `dotnet tool` commands won't work, if there is a `global.json` file
   targeting `6.0.100-alpha.1.21064.27`.

I think for now, we should make it so callers can set `dotnet: ''` and
skip many of these steps. Callers can also leave the `tools: []`
parameter blank.

If a build needs to install .NET 6 and other .NET global tools, I
think it will be able to be done in `preBuildSteps` after these
changes.

### DIFF
--- a/.ci/build.yml
+++ b/.ci/build.yml
@@ -106,11 +106,13 @@ jobs:
         inputs:
           version: ${{ parameters.dotnet }}
           performMultiLevelLookup: true
+        condition: ne('${{ parameters.dotnet }}', '')
       - pwsh: |
           dotnet tool install -g api-tools --version ${{ parameters.apiTools }}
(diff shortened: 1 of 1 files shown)


mattleibow (approved):

jonathanpeppers:
I'm just going to merge this and try it out.

It's possible I will need to change more here...

PR #1092, opened 2021-02-12
[Shopify] Remove credentials from sample code.

jonpryor:


mattleibow (approved):
Looks good from the code side. CI is unhappy though 😊

mattleibow:
I'll have a look at the ci issue. Cake was updated and they made a breaking change. Usually you could specify an empty arg, like ` --names=` but with the latest update they made that invalid. The template needs updating to first check for the value that will go there before adding the arg.

PR #1111, opened 2021-03-05
[Guava] Multi-target for .NET 6 and MonoAndroid90

jonathanpeppers:
So far ported these packages to .NET 6:

* Xamarin.Google.Guava
* Xamarin.Google.Guava.FailureAccess
* Xamarin.Google.Guava.ListenableFuture

This requires:

* Xamarin.Android from VS 16.10
* .NET 6.0.100-preview.4.21255.9
* .NET 6 Android workload
* Xamarin.Legacy.Sdk
* Updated `build.cake` to use `dotnet build`


jonathanpeppers:
Will try this again in a couple hours:
```
No matching sdk version could be found for specified version: 6.0.100-preview.4.21255.9 Kindly note the preview versions are only considered in latest version searches if Include Preview Versions checkbox is checked.
Version 6.0.100-preview.4.21255.9 could not be found in its channel, will now search in adjacent channels.
Searching for version in channel 6.0
```

Preview 4 is live here already: https://dotnet.microsoft.com/download/dotnet/6.0

But I guess the `UseDotNet` task can't see it yet.

Android/Guava/samples/GuavaSample/Future.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Google.Common.Util.Concurrent;
+using Java.Lang;
+using Java.Util.Concurrent;
+
+namespace GuavaSample
+{
+    // Just ensure this compiles
+    public class Future : Java.Lang.Object, IListenableFuture
```
jonathanpeppers, line 8: This sample isn't really that helpful, basically just an "integration test" checking a type exists.

When I look at `GuavaSample.dll` it has the same `.jar` files embedded as before:

image

jpobst:
Package looks the same, other than `MonoAndroid50` -> `MonoAndroid90`, which I assume is intended.

image

You will likely want to bump the patch version numbers so they are higher than the ones currently published on NuGet?

jpobst:
Unrelated question, we specify `net6.0-android`:
```
<TargetFrameworks>net6.0-android;monoandroid90</TargetFrameworks>
```

The TF in the NuGet is `net6.0-android30.0`.  Is that expected?

jonathanpeppers:
I didn't change any of the `monoandroid90` values, they seem the same as before?

If you leave the `30` off of `net6.0-android`, it will use a value from a generated `.targets` file:

https://github.com/xamarin/xamarin-android/blob/b36df8fbeb15bcc2bcbe239d7ce49f8ec93ad263/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs#L63-L65

Such as:
```xml
<TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">30.0</TargetPlatformVersion>
```

We could explicitly put 30 here? It will change to 31 by default at some point.

jpobst:
Maybe feels safer to explicitly use `30`?  It seems without it we run the risk of updating our XA at some point and accidentally releasing new packages that don't work for existing users.

moljac (approved):
LGTM

PR #1119, opened 2021-04-01
Io.grpc 1.29.0

moljac:
Io.grpc updates

- version bumps
- TargetFrameworks MonoAndroid90


jpobst (approved):
You might want to use a `global.json` so you don't have to manually maintain the NoTargets version number in every project:
https://github.com/xamarin/GooglePlayServicesComponents/blob/master/global.json

moljac:
@jpobst 
Yes. I planned adding `global.json` for the whole XamarinComponents repo, just to unblock some `androidx` bindings like `security` (rxjava updates and fixes needed).

PR #1147, opened 2021-05-08
master/main cleanup

moljac:
Recent changes in various repos are causing manifest build to unnecessarily build numerous folders which were recently updated (mac only). Builds fail due to old/stale cake scripts with old APIs.

Investigation was done with Matthew. It is not completely clear why this is happening.

this PR is attempt to make builds green by fixing some typos and updating cake APIs (`DeleteDirectory`).


Android/AnimatedCircleLoadingView/build.cake
```diff
@@ -43,6 +51,17 @@ Task ("clean").IsDependentOn ("clean-base").Does (() =>
 	DeleteFiles ("./externals/*.aar");
 });
 
+Task("ci")
+	//.IsDependentOn("nuget")
+	.Does 
+	(
+		() => 
+		{
+			Warning($"Not available (moljac 2021-05-08) :");
```
mattleibow, line 60: Just a note: not sure what we are supposed to do in this case... Maybe we delete the entry in the manifest? Then we can't accidentally assume it still works.

mattleibow (commented):
In all honesty, probably better to just remove these projects from the manifest and add a readme saying these old bindings are obsolete. Some have been updated last in 2015.

Shopify and OpenID probably are still good and need love as you say. Might be worth doing the same and then adding a comment in the readme under those saying that these are no longer maintained.

Android/AnimatedCircleLoadingView/build.cake
moljac: manifest entries deleted and added list to readme.md

mattleibow (approved):

Android/AnimatedCircleLoadingView/build.cake
moljac: commented out in the manifest and added list of obsoleted bindings/projects in readme

PR #1156, opened 2021-05-11
correct globbing pattern for GoogleGson assembly

moljac:
correct globbing pattern for GoogleGson.dll to fix signing failures


moljac:
Merging, because it builds - only signing fails.

Issue #1172, opened 2021-05-15
KotlinX.Coroutines buildTransitive - jar files w/o version suffix


moljac:
AndroidX samples build fails with following error:

```
       "PREFIX/X/AX/mu-window-trace/samples/BuildAll/BuildAll.sln" (Build target) (1:2) ->
       "PREFIX/X/AX/mu-window-trace/samples/BuildAll/BuildAll/BuildAll.csproj" (default target) (3:6) ->
         /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(780,3): error XACDJ7028: System.IO.FileNotFoundException: Could not find file 'PREFIX/X/AX/mu-window-trace/samples/packages/xamarin.kotlinx.coroutines.core.jvm/1.4.3/jar/kotlinx-coroutines-core-jvm.jar'. [PREFIX/X/AX/mu-window-trace/samples/BuildAll/BuildAll/BuildAll.csproj]
       /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(780,3): error XACDJ7028: File name: 'PREFIX/X/AX/mu-window-trace/samples/packages/xamarin.kotlinx.coroutines.core.jvm/1.4.3/jar/kotlinx-coroutines-core-jvm.jar' [PREFIX/X/AX/mu-window-trace/samples/BuildAll/BuildAll/BuildAll.csproj]
       /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(780,3): error XACDJ7028:   at System.IO.FileInfo.get_Length () [0x0001d] in <92218043474744ea9d64d27064c35dcb>:0  [PREFIX/X/AX/mu-window-trace/samples/BuildAll/BuildAll/BuildAll.csproj]
       /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(780,3): error XACDJ7028:   at (wrapper remoting-invoke-with-check) System.IO.FileInfo.get_Length() [PREFIX/X/AX/mu-window-trace/samples/BuildAll/BuildAll/BuildAll.csproj]
[...]

Issue #1136, opened 2021-04-28
Please update Kotlin StdLib NuGet package


humblehacker:
The latest version of the [Kotlin StdLib Nuget is 1.4.20](https://www.nuget.org/packages/Xamarin.Kotlin.StdLib/1.4.20), last updated 5 months ago. Since then, versions [1.4.21](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib/1.4.21), [1.4.30](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib/1.4.30), [1.4.31](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib/1.4.31), [1.4.32](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib/1.4.32), and [1.5.0](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib/1.5.0) of the native Android library have been released. Please consider updating this NuGet package to one or more of these available versions.

humblehacker:
@jpobst fyi, I was asked to post this on `xamarin/xamarin-android` by @Redth. Should I post future requests for Nuget updates here instead?  I have quite a few more of these requests to make.

jpobst:
Unfortunately, the code for the various Xamarin Nugets is spread all over the place.

AndroidX: https://github.com/xamarin/AndroidX
GooglePlayServices/Firebase: https://github.com/xamarin/GooglePlayServicesComponents
Facebook Components: https://github.com/xamarin/FacebookComponents

And then this repository is just kinda "everything else", which happens to include Kotlin Stdlib.  :wink: 
https://github.com/xamarin/XamarinComponents/tree/master/Android/Kotlin

tranb3r:
Updating this package could also be the opportunity to fix the following issue:
https://github.com/xamarin/XamarinComponents/issues/1017

4brunu:
I also need this.
Could you please update kotlin to version 1.5.0?

tranb3r:
@moljac 
I see you've opened new PRs for kotlin update.
Could you please fix the buildTransitive issue mentionned here #1017 and here #754 before publishing new nugets ?
Thanks.

moljac:
@tranb3r 

seems like it has already been done:

```
  <ItemGroup>
[...]

Issue #1017, opened 2020-11-13
Use buildTransitive in Kotlin bindings


tranb3r:
Kotlin bindings use .targets files.
Is there any reason not to use buildTransitive for these .targets, in order to make sure the .jar files are always propagated up to the top-level application ?

More context can be found here: https://github.com/xamarin/XamarinComponents/issues/754

moljac:
Done.

Discussed here: https://github.com/xamarin/XamarinComponents/issues/1136#issuecomment-834470622

Closing.

PR #1173, opened 2021-05-15
KotlinXCoroutines-1.4.3.1-targets-files-fixes

moljac:
refactoring + fixes:

* PackageVersion and ArtifactVersion 

* suffixes in targets files added

  details: https://github.com/xamarin/XamarinComponents/issues/1172
  
* javasources download (since version 1.4.3 there are no javadocs) 

* `JavaSourceJar` added

### DIFF
--- a/Android/KotlinXCoroutines/source/KotlinX.Coroutines.Android/KotlinX.Coroutines.Android.csproj
+++ b/Android/KotlinXCoroutines/source/KotlinX.Coroutines.Android/KotlinX.Coroutines.Android.csproj
@@ -22,7 +22,11 @@
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=2083771</PackageProjectUrl>
     <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageVersion>1.4.3</PackageVersion>
+    <PackageVersion>1.4.3.1</PackageVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <ArtifactVersion>1.4.3</ArtifactVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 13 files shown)


Android/KotlinXCoroutines/source/KotlinX.Coroutines.Android/KotlinX.Coroutines.Android.csproj
```diff
@@ -39,8 +43,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="..\..\externals\kotlinx-coroutines-android-$(PackageVersion).jar" Pack="True" PackagePath="jar\kotlinx-coroutines-android-$(PackageVersion).jar" Visible="false" />
-    <EmbeddedJar Include="..\..\externals\kotlinx-coroutines-android-$(PackageVersion).jar" Link="Jars\kotlinx-coroutines-android-$(PackageVersion).jar" />
+    <None Include="..\..\externals\kotlinx-coroutines-android-$(ArtifactVersion).jar" Pack="True" PackagePath="jar\kotlinx-coroutines-android-$(ArtifactVersion).jar" Visible="false" />
+    <EmbeddedJar Include="..\..\externals\kotlinx-coroutines-android-$(ArtifactVersion).jar" Link="Jars\kotlinx-coroutines-android-$(ArtifactVersion).jar" />
```
jpobst: What do these lines do?

It seems like they're going to add the `.jar` to the `.nupkg` twice: once embedded in the `.dll` and once as a loose `.jar` file.

moljac: I think this is for `buildTransitive` (but I am no expert on this)

https://github.com/xamarin/XamarinComponents/issues/1136

and

https://github.com/xamarin/XamarinComponents/issues/1017

mattleibow: ~It looks to be added twice, but this is probably because the accompanying .targets file is not added to the nuget - and then this was added back.~
~The .targets file should be added to build and buildTransitive~

I am a dummy, the targets file is added in the lines above. This is embedding unnecessarily.

jpobst: It seems like we shouldn't be adding `.targets` files or adding loose `.jar` files at all, but I guess we do it with all of our packages so I guess it's ok.

jpobst (approved):

PR #1189, opened 2021-05-31
datatransport-3.0.0

moljac:
datatransport-3.0.0 update for GPS-FB-MLKit 20210531

### DIFF
--- a/Android/Google.Android.DataTransport/build.cake
+++ b/Android/Google.Android.DataTransport/build.cake
@@ -10,16 +10,16 @@ var TARGET = Argument ("t", Argument ("target", "ci"));
 Dictionary<string, string> URLS_ARTIFACT_FILES= new Dictionary<string, string>()
 {
 	{
-		$"https://maven.google.com//com/google/android/datatransport/transport-api/2.2.1/transport-api-2.2.1.aar",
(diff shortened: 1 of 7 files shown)


jpobst (commented):
I realize it's unfortunate that we shipped these bad namespaces, but I think we need to be careful about whether  it's really worth fixing them:

ex:
```
### Removed Namespace Com.Google.Android.Datatransport.Backend.Cct
### New Namespace Xamarin.Google.Android.DataTransport.Backend.Cct
```

We are still dealing with users whose apps are broken from [when we fixed the `Accessibiity`](https://github.com/xamarin/AndroidX/pull/238) namespace typo: https://developercommunity.visualstudio.com/t/XamarinFirebaseMessaging-12110-does-/1434083.

Given that these are transitive dependencies for some very popular packages like `Xamarin.Firebase.Messaging` (3M downloads), it may be better for the ecosystem if we just live with the unchanged namespaces.

moljac:
I would change it (fix it).

Few reasons:

- major version change 2.x.y -> 3.0.0
- `backend.cct` is most likely not used directly from C#, but mostly as transitive dependency. OK issues could be when manually adding it, but those users are most likely advanced users if they add it manually
- yes I make mistakes.

jpobst (approved):

PR #1196, opened 2021-06-07
SignalR.Java.Client-5.0.6

4brunu:

PR #1221, opened 2021-06-21
SignalR.Java.Client-5.0.7

4brunu:
Meanwhile version 5.0.7 was released, to this PR bumps the binding to the last version


moljac (approved):

moljac:
@4brunu 

I think namespace should be: `Xamarin.AspNet.SignalR.Java.Client` to be consistent with tons of other ASP.NET packages.

let us change it ASAP before too much people use it or with next major update.

I changed it in `*.csproj` but not sure what happened with metadata. I am sure I changed it, but seems like I did not commit it. Too much work in parallel.

4brunu:
I think you changed it here https://github.com/xamarin/XamarinComponents/pull/1196/commits/ddfe22a711cd1dba15778652ca4f9c83c10eefb7

moljac:
> I think you changed it here [ddfe22a](https://github.com/xamarin/XamarinComponents/commit/ddfe22a711cd1dba15778652ca4f9c83c10eefb7)

true. Now I am wondering what happened to metadata. i changed those too.

There are very few downloads of the previous version. I think I will change namespace in matadata.

4brunu:
Since this is publish just a few days ago, so it's still fine to change it

4brunu:
When you publish the new package can't you remove the old one, to avoid noise?

moljac:
> Since this is publish just a few days ago, so it's still fine to change it

You re right. Let's do it.

> When you publish the new package can't you remove the old one, to avoid noise?

I cannot do it myself, but I will request delisting the package

PR #1222, opened 2021-06-23
Square.Retrofit2.ConverterScalars-2.5.0

moljac:
new bindings

required by internal team, but not owned by Microsoft


jpobst:
It doesn't look like CI is actually building this?

Do we need to add an entry to https://github.com/xamarin/XamarinComponents/blob/main/manifest.yaml ?

moljac:
manifest.yml updated

jpobst (approved):

PR #1245, opened 2021-07-27
TickTockAndroidTZDB-1.1.0-2020f [Depends on another PR]

4brunu:
WARNING: This PR can only be merge after https://github.com/xamarin/AndroidX/pull/345 and https://github.com/xamarin/XamarinComponents/pull/1244

Due to the Android fragmentation, there are a lot of devices with old timezone data.
So if we want to use Java 8 `java.time.*` APIs we are stuck with old timezone data that can cause bugs.
TickTock fixes this issue by providing updated timezone data, and allowing to use Java 8 `java.time.*` APIs safely.


4brunu:
Now that https://github.com/xamarin/AndroidX/pull/345 was merged and published on nuget, this PR only depends on https://github.com/xamarin/XamarinComponents/pull/1244.

PR #1260, opened 2021-08-24
[Kotlin] Ensure .targets file gets put in both /build and /buildTransitive.

jpobst:
In https://github.com/xamarin/XamarinComponents/pull/1043, we started including the `.targets` file in the `/buildTransitive` folder of the NuGet.  Unfortunately because the `ItemGroup` uses `Update` instead of `Include` we ended up overwriting the original instruction to include it in `/build`.

So our packages have only been placing the `.targets` file in `/buildTransitive` and not `/build`.  This breaks users who are still using `packages.configs`.

From https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior:
> To construct a package which allows build assets to flow transitively, package author will need to put all these build assets in /buildTransitive as well as /build folder to make the package compatible with packages.config.

This fix puts the `.targets` back in both folders.

Results:
image

### DIFF
--- a/Android/Kotlin/source/Project.cshtml
+++ b/Android/Kotlin/source/Project.cshtml
@@ -59,8 +59,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <None Update="@(Model.NuGetPackageId).targets" Pack="True" PackagePath="build\$(TargetFramework)" />
-    <None Update="@(Model.NuGetPackageId).targets" Pack="True" PackagePath="buildTransitive\$(TargetFramework)" />
+    <None Update="@(Model.NuGetPackageId).targets" Pack="True" PackagePath="build\$(TargetFramework);buildTransitive\$(TargetFramework)" />
   </ItemGroup>
 
   <ItemGroup>


Android/Kotlin/config.json
```diff
@@ -20,14 +20,14 @@
 			"groupId" : "org.jetbrains.kotlin",
 			"artifactId" : "kotlin-stdlib",
 			"version" : "1.5.20",
-			"nugetVersion" : "1.5.20",
+			"nugetVersion" : "1.5.20.1",
```
4brunu: The last version on Nuget is 1.5.21, this is because the version 1.5.20 and 1.5.21 landed in the wrong order, so the last commit was 1.5.20, but maybe you want the 1.5.21 version, I'm not sure 🙂

jpobst: Ick, I don't see anywhere in our commit history where it was set to `1.5.21` which scares me.

I just removed doing new packages in this PR, and this fix will get published with https://github.com/xamarin/XamarinComponents/pull/1259.

Thanks for the callout!

4brunu: No problem 👍

moljac (approved):
LGTM

PR #1269, opened 2021-09-06
Update ARCore to 1.25

kamszyc:
Just the bump of ARCore version to 1.25

Newest version available right now is 1.26.
I'm willing to make another pull request to update to 1.26, but I'd like to have both 1.25 and 1.26 available on NuGet. That's why I created the pull request to update to 1.25 first


moljac (approved):
LGTM

PR #1277, opened 2021-09-13
Update to multitarget net6.0-android as well

Redth:


conceptdev:
@Redth is there something simple we can do to bump this along and get a `net6.0` version of DuoSDK to try with MAUI?

PR #1278, opened 2021-09-16
[build.yml] Allow consumers to specify pools other than Azure Pipelines.

jpobst:
Currently consumers cannot specify the VM pool.  This means it always defaults to `Azure Pipelines`.  Our GPS/FB build now hits the disk space 10 GB disk space limit on Windows, so we need to use a beefier machine in a DevDiv pool.

Adding this template parameter will allow us to specify the DevDiv pool.

Reference: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#pool

### DIFF
--- a/.ci/build.yml
+++ b/.ci/build.yml
@@ -17,7 +17,7 @@ parameters:
   publishOutputSuffix: ''                                   # the artifact suffix to use when publishing the output folder
   signListPath: 'SignList.xml'                              # the path to the SignList.xml to copy into the nuget artifact for signing
   # job software version parameters
-  agentPool: 'Azure Pipelines'                              # the name of the Azure VM pool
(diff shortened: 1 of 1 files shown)


.ci/build.yml
```diff
@@ -17,6 +17,7 @@ parameters:
   publishOutputSuffix: ''                                   # the artifact suffix to use when publishing the output folder
   signListPath: 'SignList.xml'                              # the path to the SignList.xml to copy into the nuget artifact for signing
   # job software version parameters
+  agentPool: 'Azure Pipelines'                              # the name of the Azure VM pool
```
pjcollins: Maybe we should rename this `agentPoolName` to improve clarity on the callers side.

pjcollins (approved):

PR #1280, opened 2021-09-21
[Guava] Update to 29.0 and target 'net6.0-android31'.

jpobst:
Diffs are a result of removing unneeded `Resource.Designer.cs` that was added due to a bug in early .NET 6 previews.


moljac (approved):
go

PR #1286, opened 2021-10-09
Binderator suport for interpolation and variables in POM.xml files

moljac:
Binderator was not able to handle interpolation and variables like `${project.groupId}` or `${project.version}` in `pom.xml` files (mainly for dependencies.

This PR adds support for interpolation and variables

From https://github.com/xamarin/GooglePlayServicesComponents/pull/520/.

Sample `POM.xml` (`cronet-common-72.3626.96.pom`):

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">

  <modelVersion>4.0.0</modelVersion>

  <groupId>org.chromium.net</groupId>

  <artifactId>cronet-common</artifactId>

  <version>72.3626.96</version>

  <packaging>aar</packaging>

  <name>Cronet</name>

  <description>Cronet shared implementation. Not a complete implementation; apps and libraries should not depend on this target directly.</description>

  <dependencies>

    <dependency>

      <groupId>${project.groupId}</groupId>

      <artifactId>cronet-api</artifactId>

      <version>${project.version}</version>

    </dependency>

  </dependencies>

  <licenses>

    <license>

      <name>Chromium and built-in dependencies</name>

      <url>https://storage.cloud.google.com/chromium-cronet/android/72.0.3626.96/Release/cronet/LICENSE</url>

      <distribution>repo</distribution>

    </license>

  </licenses>

</project>

```

### DIFF
--- a/Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
+++ b/Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
@@ -326,6 +326,15 @@ namespace AndroidBinderator
 
 					mavenDep.Version = FixVersion(mavenDep.Version, mavenProject);
 
+					if (mavenDep.GroupId.Contains("${project.groupId}"))
+					{
+						mavenDep.GroupId = mavenDep.GroupId.Replace ("${project.groupId}", mavenProject.GroupId);
+					}
+					if (mavenDep.Version.Contains ("${project.version}"))
+					{
+						mavenDep.Version = mavenDep.Version.Replace ("${project.version}", mavenProject.Version);
+					}
+
 					var depMapping = config.MavenArtifacts.FirstOrDefault(
 						ma => !string.IsNullOrEmpty(ma.Version)
 						&& ma.GroupId == mavenDep.GroupId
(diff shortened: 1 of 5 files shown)


Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
```diff
@@ -326,6 +326,15 @@ static List<BindingProjectModel> BuildProjectModels(BindingConfig config, Dictio
 
 					mavenDep.Version = FixVersion(mavenDep.Version, mavenProject);
 
+					if (mavenDep.GroupId.Contains("${project.groupId}"))
```
jpobst: You don't need to do the `Contains` here if you don't want to, you can just call `Replace` and if nothing is found it won't change the string.

jpobst (approved):
LGTM

Util/Xamarin.AndroidBinderator/Xamarin.AndroidBinderator/Engine.cs
moljac: this was 3rd version... It was difficult to find all, but minimal places where to add this. `if` was relict of tracing and debugging unit-tests and on the real samples.

will fix it

moljac: fixed

PR #1296, opened 2021-11-01
[Bug] [Windows] Xamarin.Build.Download Fix Max Path Length Extraction

### LINKED ISSUE #1293: Xamarin.Build.Download package id validation causes build errors
I upgraded my Xamarin.Android project to Xamarin.Firebase.Storage to v120.x.x.x to take advantage of the newer firebase emulator APIs, which caused the build error "Invalid item ID firebaseappcheckinterop-16.0.0-beta01".

Related issues reported on StackOverflow:
https://stackoverflow.com/questions/68599578/xamarin-build-download-targets-invalid-id
https://stackoverflow.com/questions/69458719/i-get-the-error-invalid-item-id-firebaseappcheckinterop-16-0-0-beta01-during-b

Digging in, I've determined that the root cause is in `Xamarin.Build.Download.DownloadUtils.ParseDownloadItems(..)`, which is highly opinionated on what should be considered a valid package id. Specifically, it expects that the version component of a package id (in this case, `16.0.0-beta01`) should only use periods as separators. Since this version component uses a dash `-` before the `beta01` postfix, the id check fails.

To confirm this, I cloned the XamarinComponents repo and modified the offending method to exclude the id check:
```
public IEnumerable<XamarinBuildDownload> ParseDownloadItems (ITaskItem[] items, bool allowUnsecureUrls)
		{
			if (items == null || items.Length <= 0)
				return new List<XamarinBuildDownload> ();
			
			var results = new List<XamarinBuildDownload> ();

			foreach (var item in items) {
				var xbd = new XamarinBuildDownload ();

				xbd.Id = item.ItemSpec;
				// !!!! THIS IS CAUSING BUILDS TO FAIL
				//if (!ValidateId (xbd.Id)) {
				//	Log.LogCodedError (ErrorCodes.XbdInvalidItemId, "Invalid item ID {0}", xbd.Id);
				//	continue;
				//}
				xbd.Url = item.GetMetadata ("Url");
......
```
After packing the modified code and copying the output to my local nuget cache (replacing xamarin.build.download v0.10.0), my Xamarin.Android project compiled and ran without issue.

My recommendations for options to fix:

1. Convince Google to stop including preview dependencies in non-preview packages 🙄 
2. Modify the id-checking logic in `DownloadUtils` to permit dash `-` delimiters OR
3. Remove the id-checking logic in `DownloadUtils` entirely. Unless there's a good reason for this that isn't apparent to me (which is totally ~possible~ likely), it seems odd that Xamarin.Build.Download would assume responsibility for enforcing any specific package id/version format.

### LINKED ISSUE #1295: [Bug] [Windows] Xamarin.Build.Download fails due to Max Path length being exceeded (260 chars)
I'm using private NuGet packages for AdMob which is built from https://github.com/xamarin/GoogleApisForiOSComponents/pull/501.

On Windows the download/extract task fails for some `Tgz` files with errors similar to the below,

`
ERROR: Can not open output file : The filename, directory name, or volume label syntax is incorrect. : C:\\Users\\daniel\\AppData\\Local\\XamarinBuildDownloadCache\\FAdM-8.12.0-test\Frameworks\GoogleMobileAdsFramework-Current\GoogleMobileAds.xcframework\ios-arm64_x86_64-simulator\GoogleMobileAds.framework\Headers\Mediation\GADMediationNativeAdAssetIdentifiers.h
`

All the files that cause an error have a path length >= 260. This shouldn't be a problem in Windows as it does now have support for long paths, but you'll also notice the double slash separators (e.g. `C:\\Users\\daniel`). `Xamarin.Build.Download` is escaping any backslash in the path, which it shouldn't need to do on Windows.

My guess is that there is an old Win32 API that is trying to normalise the path & doesn't understand how to handle long paths.

dmariogatto:
Fix for issue #1295.

Simplified `AppendQuoted`, now no longer tries to escape `\` or `"` and just wraps the text argument in double quotes  (this I also believe is the same behaviour of [QuotedArgument](https://github.com/cake-build/cake/blob/dc550dd786baf28a63258fd994136a3fd266c1c5/src/Cake.Core/IO/Arguments/QuotedArgument.cs) in Cake). 

As mentioned in the above issue this was found when testing NuGets built from https://github.com/xamarin/GoogleApisForiOSComponents/pull/501 and is a blocking issue as it prevents building in VS for Windows.

New test case added `TestPathGreaterThan260Chars`.

Test have been validated on,

- [x] Windows 10 21H1
- [x] macOS 12.0.1

### DIFF
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinDownloadArchives.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinDownloadArchives.cs
@@ -287,10 +287,22 @@ namespace Xamarin.Build.Download
 			default:
 				throw new ArgumentException ("kind");
 			}
-			return new ProcessStartInfo (args.ProcessPath, args.ToString ()) {
-				WorkingDirectory = contentDir,
-				CreateNoWindow = true
-			};
+
+			ProcessStartInfo psi = null;
+			if (Platform.IsWindows)
+				psi = new ProcessStartInfo (args.ProcessPath, args.ToString ())
+				{
+					WorkingDirectory = null,
+					CreateNoWindow = true
+				};
+			else
+				psi = new ProcessStartInfo (args.ProcessPath, args.ToString ())
+				{
+					WorkingDirectory = contentDir,
+					CreateNoWindow = true
+				};
+
+			return psi;
 		}
 
 		static ProcessArgumentBuilder Build7ZipExtractionArgs (string file, string contentDir, string user7ZipPath, bool ignoreTarSymLinks, string vsInstallRoot)
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download.Tests/Test.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download.Tests/Test.cs
@@ -122,7 +122,7 @@ namespace NativeLibraryDownloaderTests
 
 			var success = BuildProject (engine, project, "_XamarinBuildDownload", log);
 
-			
+
 			AssertNoMessagesOrWarnings (log, DEFAULT_IGNORE_PATTERNS);
 			Assert.True (success);
 
@@ -332,7 +332,7 @@ namespace NativeLibraryDownloaderTests
 			var artifactXbdId = "gpsbasement-16.2.0";
 
 			var r = AndroidAarAdd(unpackDir, artifactXbdId, "https://dl.google.com/dl/android/maven2/com/google/android/gms/play-services-basement/16.2.0/play-services-basement-16.2.0.aar", true);
-			
+
 			AssertNoMessagesOrWarnings(r.logs, DEFAULT_IGNORE_PATTERNS);
 			Assert.True(r.success);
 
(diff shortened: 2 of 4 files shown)


Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinDownloadArchives.cs
```diff
@@ -287,10 +287,22 @@ ProcessStartInfo CreateExtractionArgs (string file, string contentDir, ArchiveKi
...
-			return new ProcessStartInfo (args.ProcessPath, args.ToString ()) {
-				WorkingDirectory = contentDir,
-				CreateNoWindow = true
-			};
+
+			ProcessStartInfo psi = null;
+			if (Platform.IsWindows)
+				psi = new ProcessStartInfo (args.ProcessPath, args.ToString ())
+				{
+					WorkingDirectory = null,
```
moljac, line 295: could you provide link to docs for this trick, please?

moljac:
possibly related

https://github.com/xamarin/XamarinComponents/issues/1293


at least merge together

Util/Xamarin.Build.Download/source/Xamarin.Build.Download/XamarinDownloadArchives.cs
dmariogatto: Setting to `null` will default to `System32` as the working directory, see [here](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.workingdirectory?view=net-5.0#remarks). This should be fine as no user/pass is specified & all the paths for the process and extraction are absolute. If we set it to `contentDir` it'll fail for paths over 260 chars.

moljac (approved):
LGTM

PR #1330, opened 2022-02-22
[ci] Add 'windowsImageOverride' parameter to access 1ES hardened images.

jpobst:
In order to access the 1ES hardened Windows image, we need to specify the following YAML `ImageOverride` which is not currently supported:

```yaml
  pool:
    name: $(1ESWindowsPool)
    demands: 
    - ImageOverride -equals $(1ESWindowsImage)
```

Add an `windowsImageOverride` template parameter to trigger this syntax.

### DIFF
--- a/.ci/build.yml
+++ b/.ci/build.yml
@@ -26,6 +26,7 @@ parameters:
                                                             # macOS-latest = macOS-10.15
                                                             # macOS-11 required for XCode 13.1
   windowsImage: 'windows-latest'                            # the name of the Windows VM image
+  windowsImageOverride: ''                                  # used to access 1ES hardened images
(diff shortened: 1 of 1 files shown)


jpobst:
@pjcollins Does this look correct? Unfortunately it's hard to test without being in `main` since it's used by other repos.

pjcollins (approved):
I think this looks good, but maybe we can test some trivial build against this branch somewhere to be sure?

jpobst:
I figured out how to run a build against this branch, and it seems to work:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5784521&view=results

PR #1331, opened 2022-02-23
[ci] Support for EO agent pools

mjbond-msft:
Related work item: VS #[1488096](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1488096)

Follow up to https://github.com/xamarin/XamarinComponents/pull/1330.  This PR helps ensure that the `windowsImageOverride` setting is only applied to windows jobs.  Plus ensures that the windows job will execute when `windowsImageOverride` is set and `windowsImage` is not. Finally, allows for targeting Windows pools by name and not just by the image name.

Per Executive Order (EO) builds must target secured agent pools such as the `AzurePipelines-EO` pool.  To target such a pool you need to specify the ImageOverride demand as in the following example:

pool:
&nbsp;&nbsp;name: 'AzurePipelines-EO'
&nbsp;&nbsp;demands:
&nbsp;&nbsp;- ImageOverride -equals AzurePipelinesWindows2022compliant

PR #1333, opened 2022-02-24
[ci] Fix windows conditional ImageOverride demand

mjbond-msft:
Follow up fix to the following PRs: https://github.com/xamarin/XamarinComponents/pull/1330, https://github.com/xamarin/XamarinComponents/pull/1331

Due to complications with what appears to be a Azure DevOps bug in attempting to apply a matrix variable as a agent pool demand, I've introduced the `ImageOverride` capability in a new `build.v1.yml` template.  To set the `windowsImageOverride` parameter, clients will need to update their template reference from `build.yml` to `build.v1.yml` 

Refactored the windows job out of the matrix build into its own job.  Doing so allows for the static `windowsImageOverride` condition to be successfully applied to the inclusion/exclusion of the `ImageOverride` demand. Plus copied all of the job steps to a dedicated `build.steps.v1`  template

I tried introducing a `poolDemand` variable in the matrix expecting the following to work, but it does not:
```
    pool:
      name: $(poolName)
      vmImage: $(imageName)
      demands:
      - $(poolDemand)      # Azure DevOps fails to expand $(poolDemand) and so leads to a 'invalid demand $(poolDemand)' error
```
Article related to the issue of Azure DevOps if conditions being limited to evaluating variables/parameters available at YAML compile-time and not at run-time
https://stackoverflow.com/questions/67322187/azure-devops-yaml-if-condition


mattleibow:
@mjbond-msft Can the demands not be a parameter? 

```
    pool:
      name: $(poolName)
      vmImage: $(imageName)
      demands:
        - ${{ parameters.poolDemand }}
```

jpobst (approved):

mjbond-msft:
@mattleibow - negative. The template encapsulates a loop (aka the matrix) that loops through each OS and so a variable is needed in this case.  The pool demand parameter would essentially be a constant with the specific demand being applied to each OS.  And since only a specific agent within a specific pool (AzurePipelines-EO) applied to a specific OS (windows) possesses the `ImageOverride` capability attempting to apply such a demand as a parameter across all pools/agents is going to break.  BTW: Ideally, what should work is setting the `poolDemand` (as the default value) to something that is guaranteed to exist such as the `Agent.Name` capability.  However, the apparent Azure DevOps bug/limitation involving using variables for demands prevents that

What I did find that works is replacing the matrix with a parameter array of `BuildPlatforms` similar to what Maui does
https://github.com/dotnet/maui/blob/69d30b39425998871bb89b9d8143dac24813bcf3/eng/pipelines/handlers.yml#L120

However, moving to a `BuildPlatforms` array exposed as a parameter represents an interface change to calling clients.  Since I'm moving to a new template we could break the interface, but such an interface (i.e., having to redefine the BuildPlatforms array if wanting to use something other than the default) seems more involved for clients to use.  So have chosen to keep the interface essentially the same

PR #1343, opened 2022-03-03
[ci] Target Linux EO compliant agent pool

mjbond-msft:
Related PR: https://github.com/xamarin/XamarinComponents/pull/1332

Per Executive Order (EO) migrate all jobs that currently use a Linux hosted pool to instead use the `AzurePipelines-EO` agent pool instead.

Executive Order
https://eng.ms/docs/initiatives/executive-order/executive-order-requirements/executiveorderoncybersecurity/buildinfraops

Solution made possible by https://github.com/xamarin/XamarinComponents/pull/1333. The `XamarinComponents` build leverages the same templates made available to external repos to build itself.

Test build using XamarinComponents to build itself based on these changes: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5832892&view=results

### DIFF
--- a/.ci/build.v1.yml
+++ b/.ci/build.v1.yml
@@ -120,7 +120,7 @@ jobs:
 
 - ${{ if ne(parameters.macosImage, '') }}:
   - job: macos
-    displayName: ${{ parameters.displayName }}
+    displayName: '${{ parameters.displayName }} macos'
     timeoutInMinutes: ${{ parameters.timeoutInMinutes }}
     continueOnError: ${{ eq(parameters.continueOnError, 'true') }}
     dependsOn: ${{ parameters.dependsOn }}
--- a/.ci/build.v1.yml
+++ b/.ci/build.v1.yml
@@ -24,6 +24,7 @@ parameters:
   macosAgentPoolName: 'Azure Pipelines'                     # the name of the macOS VM pool
   windowsAgentPoolName: 'Azure Pipelines'                   # the name of the Windows VM pool
   linuxImage: ''                                            # the name of the Linux VM image (linuxImage: 'Hosted Ubuntu 1604')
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -81,5 +81,5 @@ jobs:
   - ${{ if eq(variables['System.TeamProject'], 'devdiv') }}:
     - template: sign-artifacts/jobs/v2.yml@internal-templates
       parameters:
-        dependsOn: [ 'build' ]
+        dependsOn: [ 'linux', 'macos', 'windows' ]
```
jpobst: Does this require all jobs to run and succeed?  For example, we only use `windows` currently.  Will this still sign our artifacts?

mjbond-msft: Yes, but that's specific to the `XamarinComponents` build, which uses and depends on all 3 builds. In your case, you will only need `dependsOn['windows']`. I believe you should already be using this per the previous PR that added windows EO agent pool support

mjbond-msft: The changes in this PR should have no impact on your current build. If they do then let me know.

mattleibow (approved):

jpobst (approved):
We're not using the `.ci/build.v1.yml` file in AndroidX/GPS, so if this works for you then it's fine by us.

.ci/build.v1.yml
```diff
@@ -56,28 +57,76 @@ parameters:
...
+        cakeFile: ${{ parameters.cakeFile }}
+        cakeExtraArgs: ${{ parameters.cakeExtraArgs }}
+        forceBuild: ${{ parameters.forceBuild }}
+        namesFilter: ${{ parameters.namesFilter }}
+        targetsFilter: ${{ parameters.targetsFilter }}
+
 
-    displayName: ${{ parameters.displayName }}
+- ${{ if ne(parameters.macosImage, '') }}:
+  - job: macos
```
jfversluis, line 122: Is this job name needed? This is used as a [template for Xamarin.Essentials](https://github.com/xamarin/Essentials/blob/main/azure-pipelines.yml#L121) and because we're repeating it a couple of times this now breaks us because this is seen as not unique anymore

PR #1347, opened 2022-03-08
CG security fix

moljac:
Description
TensorFlow is an end-to-end open source platform for machine learning. The fix for CVE-2020-15209(https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15209) missed the case when the target shape of Reshape operator is given by the elements of a 1-D tensor. As such, the fix for the vulnerability(https://github.com/tensorflow/tensorflow/blob/9c1dc920d8ffb4893d6c9d27d1f039607b326743/tensorflow/lite/core/subgraph.cc#L1062-L1074) allowed passing a null-buffer-backed tensor with a 1D shape. The fix will be included in TensorFlow 2.5.0. We will also cherrypick this commit on TensorFlow 2.4.2, TensorFlow 2.3.3, TensorFlow 2.2.3 and TensorFlow 2.1.4, as these are also affected and still in supported range.

Usage


jpobst:
I don't think we ship this out of this repo anymore.  We have this bound now in GPS:
https://github.com/xamarin/GooglePlayServicesComponents/blob/ceebf35e3817572bbceee2e829bb2417b2dc6d63/config.json#L1357-L1365

We have already published 2.5.0 packages:
https://www.nuget.org/packages/Xamarin.TensorFlow.Lite/2.5.0.2

We probably need to get around to removing things from XC that we have moved to AndroidX/GPS.

moljac:
> I don't think we ship this out of this repo anymore. We have this bound now in GPS: https://github.com/xamarin/GooglePlayServicesComponents/blob/ceebf35e3817572bbceee2e829bb2417b2dc6d63/config.json#L1357-L1365

You are right. This is only to make CG scans happy.


> We have already published 2.5.0 packages: https://www.nuget.org/packages/Xamarin.TensorFlow.Lite/2.5.0.2

I know.

Peter and I did "fix" similar critical scan reports for Gson and some others. Besides there might be references to critical artifacts in folders which were not moved to AX and GSP-FB-MLKit repos. We just discussed. Either we kill this repo and listen to community or we (I) keep on fixing those. For now we look good with CG scan. There is tons of missing legal info, but for those we'll need scripts.

> We probably need to get around to removing things from XC that we have moved to AndroidX/GPS.

I cannot move Guava until I am sure that template set works. (And Guava had medium level security warning)

moljac:
yes today was deadline for CG issues with high priority (executive order)

jpobst:
So are we intending to release packages from this PR? Or is it simply to update the `cgmanifest.json`?

moljac:
> Or is it simply to update the `cgmanifest.json`?

just update, but not only `cgmanifest.json` (not sure if for example `*.csproj`s are scanned, so I updated those too)

jpobst (approved):

PR #1348, opened 2022-03-09
AndroidApplication check in targets

moljac:


moljac:
@jonathanpeppers 
How about when added as reference to Android Library (not bindings)?

jonathanpeppers (approved):

jonathanpeppers:
@moljac the reason to check for `$(AndroidApplication)` is so other class-library or binding projects don't bundle & redistribute these .jar files. What you would end up with is duplicate `.jar` files in that case.

PR #1352, opened 2022-03-22
[ci] Use 'dotnet tool update' instead of 'dotnet tool install' to handle tools already being installed.

jpobst:
AndroidX/GooglePlayServices switched to using a private agent pool in order to meet the EO requirements.  This pool is elastic, however agents can be reused if they become available while a job is waiting in the queue.

This means that the image may not always be fresh, and `dotnet` global tools that have been installed will already be installed.

Using `dotnet tool install` throws an error if the tool is already installed, however using `dotnet tool update` will install or update the tool without complaining.  Switch to using `update` to support agent reuse.

```
Tool 'api-tools' is already installed.
Tool 'cake.tool' is already installed.
Tool 'microsoft.dotnet.xharness.cli' is already installed.
##[error]PowerShell exited with code '1'.
```

### DIFF
--- a/.ci/build.yml
+++ b/.ci/build.yml
@@ -133,13 +133,13 @@ jobs:
       # custom init steps
       - ${{ parameters.initSteps }}
       - pwsh: |
-          dotnet tool install -g api-tools --version ${{ parameters.apiTools }}
-          dotnet tool install -g cake.tool --version ${{ parameters.cake }}
-          dotnet tool install -g Microsoft.DotNet.XHarness.CLI --version ${{ parameters.xharness }} --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json
+          dotnet tool update -g api-tools --version ${{ parameters.apiTools }}
+          dotnet tool update -g cake.tool --version ${{ parameters.cake }}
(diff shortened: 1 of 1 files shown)


.ci/build.yml
```diff
@@ -133,13 +133,13 @@ jobs:
       # custom init steps
       - ${{ parameters.initSteps }}
       - pwsh: |
-          dotnet tool install -g api-tools --version ${{ parameters.apiTools }}
-          dotnet tool install -g cake.tool --version ${{ parameters.cake }}
-          dotnet tool install -g Microsoft.DotNet.XHarness.CLI --version ${{ parameters.xharness }} --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json
+          dotnet tool update -g api-tools --version ${{ parameters.apiTools }}
+          dotnet tool update -g cake.tool --version ${{ parameters.cake }}
+          dotnet tool update -g Microsoft.DotNet.XHarness.CLI --version ${{ parameters.xharness }} --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json
```
jonathanpeppers, line 138: This seems to work even if I've uninstalled the tool:
```
> dotnet tool uninstall -g api-tools
Tool 'api-tools' (version '1.3.4') was successfully uninstalled.
> dotnet tool update -g api-tools --version 1.1.0-preview.1
You can invoke the tool using the following command: api-tools
Tool 'api-tools' (version '1.1.0-preview.1') was successfully installed.
```

jonathanpeppers (approved):

PR #1353, opened 2022-03-31
[XBD] Changing XBD MasterBeforeTargets value on .net6.0-ios builds

SotoiGhost:
With .net6.0-ios builds, executing XBD before `GetFrameworkPaths` target is too late.

At this point frameworks bundles downloaded with XBD are embedded correctly to the build but not processed nor copied to the final .app due to `_CollectBundleResources` target is executed once and way before `GetFrameworkPaths` on .net6.0.

Although `_CollectBundleResources` target is called multiple times during the build, with .net6.0 it's designed to be executed once per RID to save time.

This PR overrides the `MasterBeforeTargets` property with .net6.0 builds so stuff downloaded using XBD are embedded at a proper time.

Also, it seems that https://search.maven.org/ needs the User-Agent header to not throw a 403


Redth (approved):

dalexsoto (approved):

moljac (approved):
😄

PR #1354, opened 2022-04-01
Reinstate dynamic job name in build.v1.yml

jfversluis:
As part of #1343 the job names became hardcoded which is an issue if you use this yml file as a template and repeat it multiple times for different steps as we do [in Essentials](https://github.com/xamarin/Essentials/blob/main/azure-pipelines.yml). See an example of the error in Azure DevOps below:

image

This PR reinstates the dynamic job name again that is determined through the `name` parameter.

As seen here, this works for our build: https://github.com/xamarin/Essentials/pull/1985


jfversluis:
@jpobst seeing you were on the mentioned PR tagging you here for a potential review :)

jpobst (approved):
We don't use the `v1` version, so you're free to do whatever you want. 😁

PR #1357, opened 2022-04-04
Update build.steps.v1.yml

jfversluis:
It seems like `variables['System.JobName']` isn't really reliable as it returns `__default`. Also, because of the change in #1354 the job name is now dynamic (again).

image

Because of this jobs were being skipped in our pipeline where they shouldn't be, causing build failures. This change adds checks if the agent is running on Darwin (macOS) or Linux for better handling of these scenarios


jfversluis:
@jpobst if I could bother you for another approval please? 😬

jpobst (approved):

PR #1366, opened 2022-04-19
[Guava.ListenableFuture] Create a 9999.0 version without the .jar to match Maven.

jpobst:
### DO NOT MERGE THIS TO `main`, THIS WILL BE PUBLISHED OUT OF THE BRANCH

Creates a `9999.0.0` version of `Xamarin.Google.Guava.ListenableFuture` that does not include `guava-listenablefuture.jar`.  This version of the package will be the dependency for `Xamarin.Google.Guava`.  It will be delisted so that it does not show up for people adding an explicit reference to `Xamarin.Google.Guava.ListenableFuture`.

This mirrors the way Google/Maven have set this up:
https://mvnrepository.com/artifact/com.google.guava/listenablefuture

There are 3 scenarios we need to cover to ensure one (and only one) copy of `ListenableFuture.class` is included in an application.

- **Project (and dependencies) only references `Xamarin.Google.Guava.ListenableFuture`**
  - Project will reference the _listed_ version in NuGet (currently `1.0.0.7`)
  - `ListenableFuture.class` will be provided by `guava-listenablefuture.jar`
- **Project (and dependencies) only references `Xamarin.Google.Guava`**
  - `Xamarin.Google.Guava` will have a dependency on `Xamarin.Google.Guava.ListenableFuture` `9999.0` which does not include `guava-listenablefuture.jar`.
  - `ListenableFuture.class` will be provided by `guava.jar`
- **Project (and dependencies) references `Xamarin.Google.Guava.ListenableFuture` and `Xamarin.Google.Guava`**
  - `Xamarin.Google.Guava`'s dependency on `Xamarin.Google.Guava.ListenableFuture` `9999.0` will ensure this version is used.
  - `ListenableFuture.class` will be provided by `guava.jar`

Note `9999.0` package does not include `guava-listenablefuture.jar` or `Xamarin.Google.Guava.ListenableFuture.targets`.

image

Note that we _do_ provide managed bindings for `Xamarin.Google.Guava.ListenableFuture`, however we do not provide _any_ managed bindings for `Xamarin.Google.Guava`.  Thus the `9999.0` must still include the C# binding of `ListenableFuture` for applications that use it, since `Xamarin.Google.Guava` will not.

image

### DIFF
--- a/Android/Guava/Directory.Build.props
+++ b/Android/Guava/Directory.Build.props
@@ -5,7 +5,7 @@
     <!-- <PackageVersionSuffix Condition=" '$(BUILD_BUILDID)' != '' ">$(PackageVersionSuffix).$(BUILD_BUILDID)</PackageVersionSuffix> -->
     <GuavaNuGetVersion>31.1.0.1$(PackageVersionSuffix)</GuavaNuGetVersion>
     <GuavaFailureAccessNuGetVersion>1.0.1.6$(PackageVersionSuffix)</GuavaFailureAccessNuGetVersion>
-    <GuavaListenableFutureNuGetVersion>1.0.0.7$(PackageVersionSuffix)</GuavaListenableFutureNuGetVersion>
+    <GuavaListenableFutureNuGetVersion>9999.0</GuavaListenableFutureNuGetVersion>
     
     <!-- Opt out of C#8 features to maintain compatibility with legacy -->
     <AndroidBoundInterfacesContainConstants>false</AndroidBoundInterfacesContainConstants>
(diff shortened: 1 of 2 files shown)


jpobst:
Leaving this as a draft so it doesn't get accidentally committed.

jonathanpeppers (approved):
LGTM, would we probably just keep this on a `guava-9999` branch forever?

jpobst:
Published and unlisted:
https://www.nuget.org/packages/Xamarin.Google.Guava.ListenableFuture/9999.0.0

PR #1379, opened 2022-05-31
[ci] Add custom build steps that run after the API Diff.

jpobst:
In AndroidX https://github.com/xamarin/AndroidX/pull/558, we added a new check that ensure all new namespaces are approved.  This `postBuildStep` runs correctly, however it runs before the API Diff generator.

In the case where this new check fails and aborts the build, it would be nice to have access to the API-diffs to see the new API to review.

This PR adds a new `postDiffBuildSteps`  section where custom steps can be run after the API Diff is run.

Note: `main` for this repository seems to be broken.  This PR only affects `build.yml`, which is not used by this repository's CI.


moljac:
failing?

moljac (approved):

PR #1380, opened 2022-06-02
Allow Xamarin.Build.Download to optional bypass validate ID 

JimmyPun610:
Since invalid item id always occurs if the library is beta version, this update allow us bypass the validation so Xamarin.Build.Download will not block the build.

Issue related: 
#1293 
[Invalid item ID image-1.0.0-beta1](https://github.com/xamarin/GooglePlayServicesComponents/issues/617)
[Invalid item ID image-1.0.0-beta1](https://github.com/xamarin/GooglePlayServicesComponents/issues/621)

Accept property XamarinBuildDownloadBypassValidation from .csproj to bypass validation on downloading packages by adding these lines in csproj.

```
<Project>
....
	<Target Name="SetXamarinBuildDownloadBypassValidation" DependsOnTargets="BeforeBuild">
		<PropertyGroup>
			<XamarinBuildDownloadBypassValidation>true</XamarinBuildDownloadBypassValidation>
		</PropertyGroup>
	</Target>
....
</Project>
```

This will ignore error ErrorCodes.XbdInvalidItemId and download the beta package.

### DIFF
--- a/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/DownloadUtils.cs
+++ b/Util/Xamarin.Build.Download/source/Xamarin.Build.Download/DownloadUtils.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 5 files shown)


jpobst:
We expanded the RegEx to allow for prerelease id's so hopefully this isn't needed anymore:
https://github.com/xamarin/XamarinComponents/pull/1384

I'm going to close this, but it can be re-opened in the future if we find a case that would be better solved with a "bypass" like this.

PR #1395, opened 2022-07-28
refreshed CI settings

moljac:
- bumped several nuget versions
- matching builds for Xamarin.Android

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,6 +20,10 @@ resources:
 
 variables:
   DotNet6Version: 6.0.300
+  DotNet6Source: https://aka.ms/dotnet6/nuget/index.json
+  NuGetOrgSource: https://api.nuget.org/v3/index.json
+  XamarinDotNetWorkloadSource: workloads.json   # or url (check for recent versions - redth)
+                                                # https://aka.ms/dotnet/maui/6.0.400.json
   # matching builds
   # NOTE: there wasn't a public release of 16.11 for macOS
   LegacyXamarinAndroidPkg:  https://dl.internalx.com/vsts-devdiv/Xamarin.Android/public/6454390/6.0.4xx/0b8d593a61bb49120fdf817e8dfdbc4b33937772/xamarin.android-12.3.99.117.pkg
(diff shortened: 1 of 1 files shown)


jonathanpeppers (commented):
Does this also need to update the `android` workload?

https://github.com/xamarin/XamarinComponents/blob/df29e1fe1db5a744f3f1336e8a888b87ecfbb159/azure-pipelines.yml#L80-L83

moljac:
> Does this also need to update the `android` workload?
> 
> https://github.com/xamarin/XamarinComponents/blob/df29e1fe1db5a744f3f1336e8a888b87ecfbb159/azure-pipelines.yml#L80-L83

Not sure. But I think it makes sense to update.

moljac:
Added snippet for workload installation and update from AX and/or GPS-FB-MLKit

jonathanpeppers (commented):
Like we did in AndroidX / GPS, can we "spot check" some of the `.nupkg` files to make sure the file sizes are the same?

moljac:
> Like we did in AndroidX / GPS, can we "spot check" some of the `.nupkg` files to make sure the file sizes are the same?

this would be quite difficult for XC repo. it builds according to manifest only folder involved in the PR. So if TensorFlow was updated it would build only TensorFlow.

Maybe closing this PR w/o merging?

jonathanpeppers:
@moljac can you make a small change in Tensorflow to test at least one of them?

moljac:
> @moljac can you make a small change in Tensorflow to test at least one of them?

OK
I will bind the same version as in GPS-FB-MLKit, so that we have one parameter less to take into account.

Should I multitarget `net6.0`? 

We left it here as `monoandroid9.0`

https://github.com/xamarin/XamarinComponents/blob/main/XPlat/TensorFlow.Lite/source/Xamarin.TensorFlow.Lite.Bindings.XamarinAndroid/Xamarin.TensorFlow.Lite.Bindings.XamarinAndroid.csproj#L12

jonathanpeppers:
@moljac can you pick one that targets .NET 6? Maybe Xamarin.Android.Glide?

jonathanpeppers (approved):
This one looks fine now, but does https://github.com/xamarin/XamarinComponents/pull/1398 replace it? It seems like that one also updates the Android workload.

jpobst:
One thing to keep in mind when changing `.yaml` files for this repository is that it is set up to always use the YAML from `main`:
https://github.com/xamarin/XamarinComponents/blob/main/azure-pipelines.yml#L15-L19

Therefore it will never test your PR changes to YAML files.  (I do not know why it is set up this way 🤷‍♂️)

This commit breaks the Linux lane with:
```
Workload ID maui isn't supported on this platform.
```

As a result we can no longer publish anything out of this repository.  We need to figure out if these changes were actually needed, and if so, how can we work around it for Linux.

PR #1409, opened 2022-11-08
[ci] Enable CodeQL static analysis.

jpobst:
Enable CodeQL static analysis for [AndroidX](https://github.com/xamarin/AndroidX) and [GPS](https://github.com/xamarin/GooglePlayServicesComponents) repositories.

Note this `.yaml` is not used by _this_ repository.

Only run CodeQL on the Windows build lane.

It's hard to test this since it only runs on `main` builds, but I did run a test against AndroidX with https://github.com/xamarin/AndroidX/pull/651 to at least ensure that normal builds do not break.

### DIFF
--- a/.ci/build.yml
+++ b/.ci/build.yml
@@ -63,18 +63,23 @@ jobs:
           linux:
             poolName: ${{ parameters.linuxAgentPoolName }}
             imageName: ${{ parameters.linuxImage }}
+            runCodeQL: false
         ${{ if ne(parameters.macosImage, '') }}:
           macos:
             poolName: ${{ parameters.macosAgentPoolName }}
             imageName: ${{ parameters.macosImage }}
+            runCodeQL: false
         ${{ if ne(parameters.windowsImage, '') }}:
           windows:
             poolName: ${{ parameters.windowsAgentPoolName }}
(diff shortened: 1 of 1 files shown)


pjcollins:
@jpobst how long are the windows builds for AndroidX and GPS? I'm concerned about bloating build times and am wondering if we should be creating "compliance" pipeline for this instead.

jpobst:
The actual build step of AndroidX is about 35 minutes, GPS is about 65 minutes.

pjcollins:
You can test by adding the init/finalize tasks explicitly to the pipeline (see parts of https://github.com/xamarin/xamarin-android/pull/7522/files#diff-fd693a495110ebb53e1d597b60e531a52fc10bd9f9004e7a3c5665b2f0accda9R44). It may be worth doing this before merging to make sure things still work and that we don't make these builds way too long. Given those timings I think we may want to try to set up a separate duplicate build job that does not publish any artifacts but will run CodeQL and other compliance tasks.

jpobst:
Since it's only one `main` build every 72 hours I think we should just live with it for now.  The whole CI for these is a mess which I hope to clean up in the next few months.

pjcollins (approved):

Issue #1423, opened 2023-02-20
Unable to bind native iOS framework that uses Async/Await in iOS 14


rolfbjarne:
_From @jslew on Sat, 18 Feb 2023 16:09:54 GMT_

Having exactly the same problem as https://github.com/xamarin/xamarin-macios/issues/13605, except now trying to back-deploy to iOS 14 (not 13, which was originaly reported).  Adding a reference to `Xamarin.iOS.SwiftRuntimeSupport` does not fix this with the current toollchain.  Perhaps the package needs to be updated again?
@chamons or @rolfbjarne, any idea?

Thanks. 

### Expected Behavior
Xamarin.iOS app which references a native framework that (internally) uses swift concurrency should run without issue.

### Actual Behavior
App crashes as soon as native code using swift concurrency is run.

### Environment

<details>
<summary>Version information</summary>

```
Visual Studio Community 2022 for Mac Preview
Version 17.5 Preview (17.5 build 1766)
Installation UUID: 74672fca-f2f0-4664-b502-6ecf39ba23f4

Runtime
.NET 7.0.1 (64-bit)
Architecture: Arm64
Microsoft.macOS.Sdk 12.3.2372; git-rev-head:754abbf6a3563f6267e5717ae832b4ac25b1f2fb; git-branch:release/7.0.1xx-xcode13.3

Roslyn (Language Service)
4.5.0-3.23056.2+97881342e427ff5cdcba8f12b12ff8e6f3564431

NuGet
Version: 6.4.0.117

.NET SDK (Arm64)
SDK: /usr/local/share/dotnet/sdk/7.0.200-preview.22628.1/Sdks
SDK Versions:
	7.0.200-preview.22628.1
	7.0.100
	6.0.405
	6.0.403
	6.0.402
	6.0.401
	6.0.400
	6.0.301
	6.0.300
	6.0.202
	6.0.201
MSBuild SDKs: /Applications/Visual Studio (Preview).app/Contents/MonoBundle/MSBuild/Current/bin/Sdks

.NET SDK (x64)
SDK Versions:
	6.0.403
	6.0.402
	6.0.401
	6.0.400
	6.0.301
	6.0.300
	6.0.105
	6.0.103
	5.0.408
	5.0.406
	3.1.425
	3.1.424
	3.1.423
	3.1.422
	3.1.420
	3.1.419
	3.1.418
	3.1.417

.NET Runtime (Arm64)
[...]

PR #1431, opened 2023-07-11
Updated target to BeforeCodesign and removed non-existent parameters for swift-stdlib-tool

albertb22:
This is PR includes the changes of #1425  for #1423 and I removed non-existent parameters which caused the swift-stdlib-tool to not run.


albertb22:
@microsoft-github-policy-service agree

albertb22:
@rolfbjarne It looks like something is wrong with the Azure Pipelines because it is running for almost 3 days

dalexsoto (approved):

